PR #1403, opened 2014-10-06
Add Math polyfill

arv:
This implements the new ES6 Math functions.

The implementation is based on V8 and MDN as well as the spec.

The tests are mostly just smoke tests. I'm not testing a lot of
edge cases.


Yaffle:
Hi, if you are intersting, there are some other implementations at https://github.com/kangax/compat-table/issues/392#issuecomment-240196072 and information about inaccurate implementations.

PR #1417, opened 2014-10-15
Add support for class member variable

vicb:
I am looking for:
- feedback on the implemenation,
- how to debug (currently I'm doing make + node debug which is not the most user friendly workflow),
- what kind of test should be added.

/cc @arv

### DIFF
--- a/src/syntax/Parser.js
+++ b/src/syntax/Parser.js
@@ -222,6 +222,7 @@ import {
   PropertyMethodAssignment,
   PropertyNameAssignment,
   PropertyNameShorthand,
+  PropertyVariableDeclaration,
   RestParameter,
   ReturnStatement,
   SetAccessor,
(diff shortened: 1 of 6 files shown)


src/syntax/Parser.js
```diff
@@ -2170,6 +2171,13 @@ export class Parser {
         annotations, body);
   }
 
+  parsePropertyVariableDeclaration_(start, isStatic, name, annotations) {
+    var typeAnnotation = this.parseTypeAnnotationOpt_();
+    this.eat_(SEMI_COLON);
```
caitp: doesn't `TypeAnnotationopt` indicate that the type annotation is optional?

vicb: I think the comment is on `this.eat_(SEMI_COLON);` / ";" 
No question typeAnnotation is optional.

vicb: I think the comment is on `this.eat_(SEMI_COLON);` / ";"

caitp: oh you're right, I misread the symbol.

PR #1515, opened 2014-11-18
Interface for json

arv:


arv:
I updated this PR so that it works again.

I want to finish this work since the d.ts file is so much easier to maintain.

One approach is to land this as is and then
- [ ] Remove the .json file and have the build tools get a json object from this file.
- [ ] Update the build tools to operate directly on the tree?
- [ ] Switch to use `tval`

arv:
I'm not sure how this ever worked before. It has a chicken and egg problem. To generate the ParseTreeVisitor we need to read the d.ts file and convert that which depends on the ParseTreeVisitor :'(

I could use the `traceur.syntax.ParseTreeVisitor` which is pretty sucky.

arv:
I'm closing this for now.

PR #1549, opened 2014-12-02
Move regexpu to devDependencies, as it is not required in deps.

juliemr:


arv:
@mathiasbynens How did it include the license before?

mathiasbynens:
```bash
$ npm install traceur@0.0.72 # or any version that had regexpu as a dependency
$ cat node_modules/traceur/node_modules/regexpu/LICENSE-MIT.txt
[license text]
```

arv:
I see.

I'll have to ask some "license-expert" what the best way to solve this would be.

PR #1593, opened 2014-12-16
Add support for initializing member variables

vicb:
The implementation follow TS rules:
- when the memberVariables option is enabled, the constructor must call the super constructor,
- when the memberVariables option is enabled and a class has initialized instance variables, the call to the super constructor must be the first statement in the class own constructor,
- instance variables initializers are evaluated when the class is instantiated.

### DIFF
--- a/src/codegeneration/ClassTransformer.js
+++ b/src/codegeneration/ClassTransformer.js
@@ -25,10 +25,13 @@ import {
   SetAccessor
 } from '../syntax/trees/ParseTrees.js';
 import {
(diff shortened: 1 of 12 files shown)


vicb:
@arv I have rebased the PR on latest master. It should be ready to get in. Could you have a final look ? Thanks.

arv:
LGTM

arv:
Rebased and merged.

vicb:
Thanks for you help on this PR.

PR #1608, opened 2014-12-23
Support composing with input source maps

crisptrutski:
Adds option `.inputSourceMap`; when set to a SourceMap object, the output sourcemap is composed with the inputSourceMap.

### DIFF
--- a/src/Compiler.js
+++ b/src/Compiler.js
@@ -241,8 +241,16 @@ export class Compiler {
   }
 
   getSourceMap() {
-    if (this.sourceMapConfiguration_)
-      return this.sourceMapConfiguration_.sourceMapGenerator.toString();
+    if (this.sourceMapConfiguration_) {
+      var sourceMap = this.sourceMapConfiguration_.sourceMapGenerator.toString();
+      var inputSourceMap = this.sourceMapConfiguration_.inputSourceMap;
+      if (inputSourceMap) {
+        var generator = SourceMapGenerator.fromSourceMap(new SourceMapConsumer(sourceMap));
+        generator.applySourceMap(new SourceMapConsumer(inputSourceMap));
+        sourceMap = generator.toJSON();
+      }
+      return sourceMap;
+    }
   }
 
   get sourceMapInfo() {
@@ -282,12 +290,6 @@ export class Compiler {
       var sourceMappingURL =
           this.sourceMappingURL(outputName || sourceURL || 'unnamed.js');
       var sourceMap = this.getSourceMap();
-      var inputSourceMap = this.sourceMapConfiguration_.inputSourceMap;
-      if (inputSourceMap) {
-        var generator = SourceMapGenerator.fromSourceMap(new SourceMapConsumer(sourceMap));
-        generator.applySourceMap(new SourceMapConsumer(inputSourceMap));
-        sourceMap = generator.toJSON();
-      }
       compiledCode += '\n//# sourceMappingURL=' + sourceMappingURL + '\n';
       // The source map info for in-memory maps
       this.sourceMapInfo_ = {
--- a/src/Compiler.js
+++ b/src/Compiler.js
@@ -28,8 +28,10 @@ import {
 
 import {ParseTreeMapWriter} from './outputgeneration/ParseTreeMapWriter.js';
 import {ParseTreeWriter} from './outputgeneration/ParseTreeWriter.js';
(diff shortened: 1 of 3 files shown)


crisptrutski:
@johnjbarton added test, and tweaked some edge cases. 

Guessing you'll also want a documented CLI option, but new IO to load the source map file, and until `addOptions` moves from `src/Options.js` to `src/node.js` that seems problematic. Skip for now?

src/Compiler.js
```diff
@@ -28,6 +28,7 @@ import {
 
 import {ParseTreeMapWriter} from './outputgeneration/ParseTreeMapWriter.js';
 import {ParseTreeWriter} from './outputgeneration/ParseTreeWriter.js';
+import {SourceMapConsumer} from './outputgeneration/SourceMapIntegration.js';
```
arv:
```js
import {
  SourceMapConsumer,
  SourceMapGenerator
} from './outputgeneration/SourceMapIntegration.js';
```

test/unit/codegeneration/SourceMap.js
```diff
@@ -91,6 +90,65 @@ suite('SourceMap.js', function() {
...
   });
 
+  test('SourceMap with input map', function() {
+    var src = 'function foo() { return 5; }\nvar \nf\n=\n5\n;\n';
+    var srcLines = src.split('\n');
+    var filename = 'sourceMapThis.js';
+
+    // maps every character one column right and two lines down
+    var g = new SourceMapGenerator({file: 'sourceMapThis.js'});
+    for (var i=0; i<srcLines.length; i++) {
```
arv: ws after = and <

test/unit/codegeneration/SourceMap.js
```diff
@@ -91,6 +90,65 @@ suite('SourceMap.js', function() {
...
+        g.addMapping({
+          generated: {line: i + 1, column: j},
+          original: {line: i + 3, column: j + 1},
+          source: 'priorSource.js'
+        });
+      }
+    }
+    var inputMap = g.toJSON();
+
+    var compiler = new Compiler({sourceMaps: 'file',
```
arv: Indent like this

```js
 var compiler = new Compiler({
  sourceMaps: 'file',
  script: true,
  inputSourceMap: inputMap
});
```

test/unit/codegeneration/SourceMap.js
```diff
@@ -91,6 +90,65 @@ suite('SourceMap.js', function() {
...
+      // function foo() { return 5; >}<
+      {generated: {line: 3, column: 0}, original: {line: 3, column: 28}},
+      {generated: {line: 4, column: 0}, original: {line: 5, column: 1}},
+      {generated: {line: 5, column: 1}, original: {line: 8, column: 1}}
+    ];
+
+    testcases.forEach(function(testcase, caseNumber) {
+      var actual = consumer.originalPositionFor(testcase.generated);
+      var shouldBeTrue = actual.line === testcase.original.line;
+      assert.isTrue(shouldBeTrue, caseNumber + ' Line mismatch ' + actual.line);
```
arv: assert.equal or assert.strictEqual and you can skip the _message_.

arv:
LGTM with nits

crisptrutski:
@arv thanks for the feedback. Have also rebased on master for clean merge

guybedford:
:+1: this would be amazing for the next SystemJS release, allowing JSX -> ES6 plugins with source maps.

test/unit/codegeneration/SourceMap.js
```diff
@@ -83,12 +82,8 @@ suite('SourceMap.js', function() {
 
     testcases.forEach(function(testcase, caseNumber) {
       var actual = consumer.originalPositionFor(testcase.generated);
-      var shouldBeTrue = actual.line === testcase.original.line;
-      assert.isTrue(shouldBeTrue, caseNumber + ' Line mismatch ' + actual.line);
-      var expected = testcase.original.column;
-      shouldBeTrue = actual.column === expected;
-      assert.isTrue(shouldBeTrue,
-          caseNumber + ' Column mismatch ' + actual.column + ' vs ' + expected);
+      assert.strictEqual(actual.line, testcase.original.line);
```
arv: FYI, we usually use `assert.equal` unless we really need strict since it reads better. (I do believe chai made the wrong choice here though.)

arv:
LGTM, landing

PR #1618, opened 2015-01-08
Fix parsing of ArrowFunction (no LineTerminator allowed)

hzr:
This should not be parsed as an ArrowFunction:
```
x
=>1
```

I noticed over at http://kangax.github.io/compat-table/es6/#arrow_functions that Traceur doesn't handle this correctly (in all cases), but hopefully it does now.


arv:
LGTM

PR #1629, opened 2015-01-13
Remove old class runtime functions

### LINKED ISSUE #1425: Remove superCall and defaultSuperCall from the runtime
superCall and defaultSuperCall can be removed once we have pushed to nmp

arv:
Fixes #1425


arv:
@johnjbarton TBR

PR #1643, opened 2015-01-17
Change annotations to annotate

### LINKED ISSUE #1642: Traceur/Atscript - annotations vs annotate
When Traceur compiles annotations, they become a .annotations property on the object they annotate. However, the AtScript specification, or at least the AtScript Primer, which is the top Google hit when you search for AtScript specification, says the property should be called .annotate:
https://docs.google.com/document/d/11YUzC-1d0V1-Q3V0fQ7KSit97HnZoKVygDxpWzEYW0U/mobilebasic?viewopt=127

Should these be brought into sync in some fashion? Which is correct for AtScript? It's very confusing.

Also, it looks like TypeScript is looking at Annotations based off the .annotate property mentioned in the AtScript primer. https://github.com/Microsoft/TypeScript/issues/1557

hannahhoward:
Brings name of annotations metadata in line with AtScript spec by changing name to 'annotate'.

resolve #1642

### DIFF
--- a/src/codegeneration/AnnotationsTransformer.js
+++ b/src/codegeneration/AnnotationsTransformer.js
@@ -293,7 +293,7 @@ class AnnotationsScope {
       annotations = this.transformAnnotations_(annotations);
       if (annotations.length > 0) {
         metadataStatements.push(this.createDefinePropertyStatement_(target,
-            'annotations', createArrayLiteralExpression(annotations)));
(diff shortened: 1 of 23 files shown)


hannahhoward:
signed CLA.

arv:
I'm going to close this and the bug until I hear from @mhevery or @IgorMinar.

PR #1656, opened 2015-01-26
Avoid setter variable collissions

guybedford:
Because the variable name used in the setters is `m`, this will collide with any import bindings called `m`, a slight oversight on my part.

I've updated with a better name to avoid these collisions - `$__m`.


arv:
LGTM

PR #1665, opened 2015-01-29
Fix incorrect URL

kaleb:


UltCombo:
This probably explains why the Wiki `GettingStarted` page keeps getting re-created by random people. Nice work!

arv:
Thanks.

I see that you have signed the CLA. However, we need to add an entry to AUTHORS as well. Since the AUTHORS file is on master we will need to do two PR. Sorry about that.

I can also just make the change to index.html if you do not are about getting the commit?

johnjbarton:
The patch needs to be applied to master then pushed to gh-pages when we
publish.

arv:
True. I thought that we only had index.html on the gh-pages branch.

kaleb:
I don't need to be added to AUTHORS.

johnjbarton:
Thanks the fix is in the repo and will be pushed on the next publish.  See #1679

Issue #1673, opened 2015-01-30
Async generators


mnieper:
Has https://github.com/jhusain/asyncgenerator progressed far enough so that one may implement at least the fundamental bits of it in Traceur?

arv:
Yeah. It seems like it would be really useful to have a prototype of this in Traceur (under a flag of course).

PR #1684, opened 2015-02-02
For-on loops (part of ES7's async generators)

mnieper:
This is my first try to implement parts of the async generators proposal at https://github.com/jhusain/asyncgenerator; see also #1673. This pull request adds a new option `--for-on` that allows asynchronous for-on-loops inside async functions. At the moment, only the basic cases are handled, so this pull request should not be merged yet. But it can be used to discuss implementation details.

As there is no formal spec for async generators and for-on loops yet, the hope is that this implementation will finally allow to shape such a spec. During this process, the implementation will likely have to be adjusted.

### DIFF
--- a/src/Options.js
+++ b/src/Options.js
@@ -37,6 +37,7 @@ export var optionsV01 = enumerableOnlyObject({
   destructuring: true,
   exponentiation: false,
   forOf: true,
+  forOn: false,
   freeVariableChecker: false,
   generatorComprehension: false,
   generators: true,
@@ -139,12 +140,12 @@ addFeatureOption('annotations', EXPERIMENTAL);
 addFeatureOption('arrayComprehension', EXPERIMENTAL); // 11.4.1.2
 addFeatureOption('asyncFunctions', EXPERIMENTAL);
 addFeatureOption('exponentiation', EXPERIMENTAL);
+addFeatureOption('forOn', EXPERIMENTAL);
 addFeatureOption('generatorComprehension', EXPERIMENTAL);
+addFeatureOption('memberVariables', EXPERIMENTAL);
 addFeatureOption('require', EXPERIMENTAL);
 addFeatureOption('symbols', EXPERIMENTAL);
 addFeatureOption('types', EXPERIMENTAL);
-addFeatureOption('memberVariables', EXPERIMENTAL);
-
 
 var transformOptionsPrototype = {};
 
--- a/src/codegeneration/ForOnTransformer.js
+++ b/src/codegeneration/ForOnTransformer.js
@@ -1,0 +1,92 @@
+// Copyright 2015 Traceur Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {
+  FOR_ON_STATEMENT,
+  LABELLED_STATEMENT
+} from '../syntax/trees/ParseTreeType.js';
+import {TempVarTransformer} from './TempVarTransformer.js';
+import {InnerForOnTransformer} from './InnerForOnTransformer.js';
+import {
+  createIdentifierExpression as id,
+  createVariableStatement
+} from './ParseTreeFactory.js';
+
+/**
+ * Transforms for-on loops into calls to async functions
+ *
+ * Example:
+ *
+ * for (let i on o) {
+ *   if (continue) continue;
+ *   if (break) break;
+ *   if (return) return;
+ * }
+ *
+ * Becomes:
+ *
+ * do {
+ *   var $result = undefined;
+ *   let i;
+ *   await $traceurRuntime.observableForEach(o, async function ($value) {
+ *     $observer = this;
+ *     i = $value;
+ *     if (continue) return;
+ *     if (break) { $result = 0; $observer.return(); return; }
+ *     if (return) { $result = {v: undefined}; $observer.return(); return; }
+ *   });
+ *   switch ($result) {
+ *   case 0:
+ *     continue;
+ *   case undefined:
+ *     continue;
+ *   default:
+ *     return $result.v;
+ *   }
+ * } while (false)
+ */
+
+/**
+ * Desugars for-on statement.
+ */
+export class ForOnTransformer extends TempVarTransformer {
+  /**
+   * @param {ForOnStatement} original
+   * @return {ParseTree}
+   */
+  transformForOnStatement(original) {
+    return this.transformForOnStatement_(original, []);
+  }
+
+  transformForOnStatement_(original, labelSet) {
+    return InnerForOnTransformer.transform(this,
+        super.transformForOnStatement(original), labelSet);
+  }
+
+  // keep track of labels in the tree
+  transformLabelledStatement(tree) {
+    var labelSet = [tree];
+    var statement;
+    for (statement = tree.statement; statement.type === LABELLED_STATEMENT;
+        statement = statement.statement) {
+      labelSet.push(statement);
+    }
+    if (statement.type !== FOR_ON_STATEMENT) {
+      return super.transformLabelledStatement(tree);
+      // Slightly inefficient in the (unlikely) case of more than one label
+    }
+    return this.transformForOnStatement_(statement, labelSet);
+  }
+}
+
--- a/src/codegeneration/InnerForOnTransformer.js
+++ b/src/codegeneration/InnerForOnTransformer.js
@@ -1,0 +1,234 @@
+// Copyright 2015 Traceur Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the 'License');
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an 'AS IS' BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {ParseTreeTransformer} from './ParseTreeTransformer.js';
+import alphaRenameThisAndArguments from './alphaRenameThisAndArguments.js';
+import {
+  parseStatement,
+  parseStatements
+} from './PlaceholderParser.js';
+import {
+  AnonBlock,
+  Block,
+  BreakStatement,
+  ContinueStatement,
+  LabelledStatement,
+  ReturnStatement
+} from '../syntax/trees/ParseTrees.js';
+import {
+  createArgumentList,
+  createAssignmentStatement,
+  createAssignmentExpression,
+  createCaseClause,
+  createDefaultClause,
+  createIdentifierExpression as id,
+  createNumberLiteral,
+  createSwitchStatement,
+  createThisExpression,
+  createVariableDeclaration,
+  createVariableDeclarationList,
+  createVariableStatement,
+  createVoid0
+} from './ParseTreeFactory.js';
+import {ARGUMENTS} from '../syntax/PredefinedName.js';
+import {VAR} from '../syntax/TokenType.js';
+import {VARIABLE_DECLARATION_LIST} from '../syntax/trees/ParseTreeType.js';
+
+export class InnerForOnTransformer extends ParseTreeTransformer {
+  // TODO: This class has considerable overlap with
+  // FnExtractAbruptCompletions. The common code should really be refactored
+  // into an abstract base class.
+  
+  constructor(tempIdGenerator, labelSet) {
+    this.idGenerator_ = tempIdGenerator;
+    this.inLoop_ = 0;
+    this.inBreakble_ = 0;
+    this.variableDeclarations_ = [];
+    this.extractedStatements_ = [];
+    this.labelSet_ = labelSet;
+    this.labelledStatements_ = {};
+    this.observer_ = id(this.idGenerator_.getTempIdentifier());
+    this.result_ = id(this.idGenerator_.getTempIdentifier());
+    this.parentLabels_ = Object.create(null);
+    this.labelSet_.forEach((tree) => {
+      this.parentLabels_[tree.name.value] = true;
+    });
+  }
+
+  transform(tree) {
+    var value = id(this.idGenerator_.getTempIdentifier());
+    var assignment;
+    if (tree.initializer.type === VARIABLE_DECLARATION_LIST) {
+      // {var,let,const} initializer = $value;
+      assignment = createVariableStatement(
+          tree.initializer.declarationType,
+          tree.initializer.declarations[0].lvalue, value);
+    } else {
+      assignment = parseStatement `
+          ${tree.initializer} = ${value};`;
+    }
+
+    var body = new Block(tree.body.location,
+        [assignment, ...tree.body.statements]);
+    body = this.transformAny(body);
+    body = alphaRenameThisAndArguments(this, body);
+
+    // var $result = undefined
+    this.variableDeclarations_.push(
+        createVariableDeclaration(this.result_, createVoid0()));
+
+    var caseClauses = this.extractedStatements_.map(
+        (statement, index) => {
+          return createCaseClause(createNumberLiteral(index), [statement])
+        });
+    caseClauses.push(createCaseClause(createVoid0(), [
+        new ContinueStatement(null, null)]));
+    caseClauses.push(createDefaultClause(parseStatements `
+        return ${this.result_}.v;`));
+
+    var switchStatement = createSwitchStatement(this.result_, caseClauses);
+    var statement = parseStatement `
+        do {
+          ${createVariableStatement(
+            createVariableDeclarationList(VAR, this.variableDeclarations_))}
+            await $traceurRuntime.observeForEach(
+              ${tree.observable}[$traceurRuntime.toProperty(Symbol.observer)].
+                  bind(${tree.observable}),
+              async function (${value}) {
+                var ${this.observer_} = this;
+                ${body}
+              });
+          ${switchStatement}
+        } while (false);`;
+
+    var labelledStatement;
+    while (labelledStatement = this.labelSet_.pop()) {
+      statement = new LabelledStatement(labelledStatement.location,
+      labelledStatement.name, statement);
+    }
+
+    return statement;
+  }
+
+  // alphaRenameThisAndArguments
+  addTempVarForArguments() {
+    var tmpVarName = this.idGenerator_.generateUniqueIdentifier();
+    this.variableDeclarations_.push(createVariableDeclaration(
+        tmpVarName, id(ARGUMENTS)));
+    return tmpVarName;
+  }
+  // alphaRenameThisAndArguments
+  addTempVarForThis() {
+    var tmpVarName = this.idGenerator_.generateUniqueIdentifier();
+    this.variableDeclarations_.push(createVariableDeclaration(
+        tmpVarName, createThisExpression()));
+    return tmpVarName;
+  }
+
+  transformAny(tree) {
+    if (tree) {
+      if (tree.isBreakableStatement()) this.inBreakble_++;
+      if (tree.isIterationStatement()) this.inLoop_++;
+      tree = super.transformAny(tree);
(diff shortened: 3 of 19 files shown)


src/codegeneration/ForOnTransformer.js
```diff
@@ -0,0 +1,70 @@
...
+  transformForOnStatement(original) {
+    var tree = super.transformForOnStatement(original);
+    var observer = id(this.getTempIdentifier());
+    var result = id(this.getTempIdentifier());
+    var resolve = id(this.getTempIdentifier());
+    var reject = id(this.getTempIdentifier());
+
+    var assignment;
+    if (tree.initializer.type === VARIABLE_DECLARATION_LIST) {
+      // {var,let} initializer = $result.value;
```
arv: A rhs is not allowed here.

src/codegeneration/ForOnTransformer.js
```diff
@@ -0,0 +1,70 @@
...
+      assignment = createVariableStatement(
+          tree.initializer.declarationType,
+          tree.initializer.declarations[0].lvalue, result);
+    } else {
+      assignment = parseStatement `${tree.initializer} = ${result};`;
+    }
+
+    // TODO (mnieper) Handle non-local control flow (break, continue, return)
```
arv: TODO(mnieper): Handle

Consistency makes it easier to grep for these

src/syntax/Parser.js
```diff
@@ -1412,6 +1429,23 @@ export class Parser {
                               collection, body);
   }
 
+  // The for-on Statement
+  // for  (  { let | var }  identifier  on  expression  )  statement
```
arv: Why no const?

test/feature/AsyncGenerators/ForOn.js
```diff
@@ -0,0 +1,31 @@
...
+  await observer.next(2);
+  await observer.next(3);
+  observer.return(4);
+});
+
+var count = 0;
+var i;
+(async function () {
+  for (i on o1) {
+    count += i;
```
arv: This could use some more asserts. What is `this` here for example.

arv:
I think this is a good approach.

mnieper:
I am going to update the transformer in src/codegeneration/ForOnTransformer.js in a minute. Without employing techniques as in FnExtractAbruptCompletions, non-local control flow cannot be handled.

Fortunately, all the code is running in an async context, so I can switch back and forth between observable and loop code via promises.

src/codegeneration/ForOnTransformer.js
mnieper: Apart from the fact that the comment should be `initializer = $result`, what do you mean with your comment? Which rhs?

(This part was modelled after ForOfTransformer.)

src/syntax/Parser.js
mnieper: Does it make sense? The identifier is assigned the values to be observed. (See also `parseForOfStatement_` in the same file.)

mnieper:
The new commit includes the promised new transformer.

src/codegeneration/ForOnTransformer.js
```diff
@@ -0,0 +1,78 @@
...
+    return parseStatement `
+        for (var ${loopLock} = $traceurRuntime.createDeferred(),
+                 ${observerLock} = $traceurRuntime.createDeferred(),
+                 ${observer} = ${tree.observable}[$traceurRuntime.toProperty(Symbol.observer)]({
+                     next: async function(v) {
+                       ${loopLock}.resolve({value: v, done: false});
+                       ${observerLock} = $traceurRuntime.createDeferred();
+                       return ${observerLock}.promise;
+                     },
+                     throw: function (e) {
```
arv: no ws between function and (

src/runtime/async.js
```diff
@@ -0,0 +1,27 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+function Deferred() {
+  this.promise = new Promise(function (resolve, reject) {
+    this.resolve = resolve;
+    this.reject = reject;    
+  }.bind(this));
```
arv: You can use an arrow function here.

```js
  this.promise = new Promise((resolve, reject) => {
    this.resolve = resolve;
    this.reject = reject;    
  });
```

src/syntax/Parser.js
```diff
@@ -1394,6 +1406,11 @@ export class Parser {
     return type === IDENTIFIER && this.options_.forOf &&
         this.peekToken_().value === OF;
   }
+  
+  peekOn_(type) {
+    return type === IDENTIFIER && this.options_.forOn &&
+        this.peekToken_().value === ON;
```
arv: Maybe we should add peekPredefinedString_?

src/syntax/Parser.js
arv: It makes sense. ES6 allows const bindings here.

The comment for parseForOfStatement is just out of date.

src/codegeneration/ForOnTransformer.js
arv: In ES6 the loop initializer, if being a variable declaration may not have an initializer RHS.

```js
for (var x = 1 in {}) {}  // SyntaxError
for (var x = 1 of []) {}  // SyntaxError
```

src/syntax/Parser.js
arv: The const binding is important in cases like:

```js
for (const x of [0, 1]) {
  x = 2;  // TypeError
}
```

src/codegeneration/ForOnTransformer.js
mnieper: The parser won't forward this to `transformFor[On|Of]Statement`. The rhs here goes into the assignment at the top of the loop's body.

mnieper:
The for-on loop should now work as described in https://github.com/jhusain/asyncgenerator. Have taken care of your comments above. Thanks a lot for reviewing, @arv!

src/codegeneration/ForOnTransformer.js
```diff
@@ -0,0 +1,124 @@
...
+          ${tree.body}
+      }`;
+
+    while (labelledStatement = labelSet.pop()) {
+      innerStatement = new LabelledStatement(labelledStatement.location,
+      labelledStatement.name, innerStatement);
+    }
+
+    return new AnonBlock(null, parseStatements `
+        {
```
arv: Anon block should not be wrapped in a block.

test/feature/AsyncGenerators/ForOn.js
```diff
@@ -0,0 +1,98 @@
...
+  var decoratedGenerator = Object.create(generator);
+  decoratedGenerator.next = function(v) {
+    var result = generator.next(v);
+    if (result && result.done) {
+      onDone();
+    }
+    return result;
+  };
+
+  ["throw","return"].forEach(method => {
```
arv: What is the end user story here? I assume we don't expect anyone to write code like this?

mnieper: LOL

The end user story is pretty much here: https://github.com/jhusain/asyncgenerator

arv: I reread that before starting to look at this WIP. Just to be clear. Once async generators are implemented no one has to write this decorate function again?

mnieper: I don't think so. For async generators, this will only go into the transpilation. All other observables should be implementable by writing async generator functions.

mnieper: P.S.: As I haven't implemented async generators yet, I had to code everything by hand in the test.

mnieper:
This is not considered WIP anymore (assuming there are no outstanding errors anymore) and    implements for-on loop as "specced" at https://github.com/jhusain/asyncgenerator.

The test test/feature/AsyncGenerators/ForOn.module.js hopefully shows how one can work with for-on loops in asynchronous code.

Note: This PR includes the fixes of https://github.com/google/traceur-compiler/pull/1697 as it relies on them.

Please review. Thx.

Once this is accepted and working, I will start working on the other part of the proposal, async generators that produce for for-on loops consume.

mnieper:
Merge conflicts due to https://github.com/google/traceur-compiler/commit/0ada3b733d873d8d714853cd5a46d394e0c074cb. Resolving...

src/Options.js
```diff
@@ -144,7 +145,7 @@ addFeatureOption('require', EXPERIMENTAL);
 addFeatureOption('symbols', EXPERIMENTAL);
 addFeatureOption('types', EXPERIMENTAL);
 addFeatureOption('memberVariables', EXPERIMENTAL);
-
+addFeatureOption('forOn', EXPERIMENTAL);
```
arv: keep alphabetical order (I know it got out of sync because I am the only one that is obsessive enough about this)

mnieper:
Merge conflict resolved. Options (including memberVariables!) ordered.

src/codegeneration/InnerForOnTransformer.js
```diff
@@ -0,0 +1,230 @@
...
+    this.inBreakble_ = 0;
+    this.variableDeclarations_ = [];
+    this.extractedStatements_ = [];
+    this.labelSet_ = labelSet;
+    this.labelledStatements_ = {};
+    this.observer_ = id(this.idGenerator_.generateUniqueIdentifier());
+    this.result_ = id(this.idGenerator_.generateUniqueIdentifier());
+    this.parentLabels_ = {};
+    this.labelSet_.forEach((tree) =>
+      this.parentLabels_[tree.name.value] = true
```
arv: We use {} (and return in this case) if the arrow does not fit on one line.

zenparsing:
@mnieper Do you have any tests for the case where the `for-on` loop contains an `await` expression?  What does your transformation do in that case?

src/codegeneration/InnerForOnTransformer.js
```diff
@@ -0,0 +1,230 @@
...
+
+  constructor(idGenerator, labelSet) {
+    this.idGenerator_ = idGenerator;
+    this.inLoop_ = 0;
+    this.inBreakble_ = 0;
+    this.variableDeclarations_ = [];
+    this.extractedStatements_ = [];
+    this.labelSet_ = labelSet;
+    this.labelledStatements_ = {};
+    this.observer_ = id(this.idGenerator_.generateUniqueIdentifier());
```
arv: There is a problem with `generateUniqueIdentifier()`. It never releases the temp var name, making the counter get large. We usually use the TempVarTransformer and its method to release the temp var names.

src/codegeneration/InnerForOnTransformer.js
```diff
@@ -0,0 +1,230 @@
...
+    this.labelSet_.forEach((tree) =>
+      this.parentLabels_[tree.name.value] = true
+    );
+  }
+
+  transform(tree) {
+    var value = id(this.idGenerator_.generateUniqueIdentifier());
+    var assignment;
+    if (tree.initializer.type === VARIABLE_DECLARATION_LIST) {
+      // {var,let} initializer = $value;
```
arv: // {var,let,const} initializer = $value;

src/codegeneration/InnerForOnTransformer.js
```diff
@@ -0,0 +1,230 @@
...
+    var body = new Block(tree.body.location, [assignment].concat(tree.body.statements));
+    var body = this.transformAny(body);
+    body = alphaRenameThisAndArguments(this, body);
+
+    // var $result = undefined
+    this.variableDeclarations_.push(
+        createVariableDeclaration(this.result_, createVoid0()));
+
+    var caseClauses = this.extractedStatements_.map(
+        (statement, index) => createCaseClause(
```
arv: {} and return

mnieper:
@zenparsing See the last test here:

https://github.com/mnieper/traceur-compiler/blob/for-on-loop/test/feature/AsyncGenerators/ForOn.module.js

The loop body is transformed into an async function.

src/codegeneration/InnerForOnTransformer.js
mnieper: Should be `generateTempIdentifier()`. The `this.idGenerator_` is initialized from a `TempVarTransformer`.

src/codegeneration/InnerForOnTransformer.js
```diff
@@ -0,0 +1,230 @@
...
+            createVariableDeclarationList(VAR, this.variableDeclarations_))}
+            await $traceurRuntime.observeForEach(
+              ${tree.observable}[$traceurRuntime.toProperty(Symbol.observer)].
+                  bind(${tree.observable}),
+              async function (${value}) {
+                var ${this.observer_} = this;
+                ${body}
+              });
+          ${switchStatement}
+        } while (false);`;
```
arv: The comment in ForOnTransformer says `while (true)`

mnieper: That would be bad. Thx.

src/codegeneration/InnerForOnTransformer.js
```diff
@@ -0,0 +1,230 @@
...
+          ${createVariableStatement(
+            createVariableDeclarationList(VAR, this.variableDeclarations_))}
+            await $traceurRuntime.observeForEach(
+              ${tree.observable}[$traceurRuntime.toProperty(Symbol.observer)].
+                  bind(${tree.observable}),
+              async function (${value}) {
+                var ${this.observer_} = this;
+                ${body}
+              });
+          ${switchStatement}
```
arv: We should really extract this part since we do almost exactly the same thing in the block binding transformer.

Add TODO and/or file bug.

arv: In other words, the work should not happen in this PR but we should clean this up in the future.

mnieper: Full ack. Adding TODO. At the beginning, I tried to base my work on the block binding transformer but it soon diverged too much. I think both transformers should get a common base at some point in the future that handles abrupt completions while being flexible about them (for example, in my transformer a local `continue` does not get extracted).

src/codegeneration/InnerForOnTransformer.js
```diff
@@ -0,0 +1,230 @@
...
+export class InnerForOnTransformer extends ParseTreeTransformer {
+
+  constructor(idGenerator, labelSet) {
+    this.idGenerator_ = idGenerator;
+    this.inLoop_ = 0;
+    this.inBreakble_ = 0;
+    this.variableDeclarations_ = [];
+    this.extractedStatements_ = [];
+    this.labelSet_ = labelSet;
+    this.labelledStatements_ = {};
```
arv: This needs to be `Object.create(null)` or we will break with label names that conflict with property names on Object.prototype.

src/runtime/async.js
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+function observeForEach(observe, next) {
+  return new Promise((resolve, reject) => {
+    var generator = observe({
+      next(value) {
+        return next.call(generator, value);
+      },
+      throw (error) {
```
arv: no ws between method name and parameter list.

test/feature/AsyncGenerators/ForOn.module.js
```diff
@@ -0,0 +1,76 @@
+// Async.
+// Options: --async-functions --for-on --symbols
+
+import {AsyncGeneratorFunction} from './resources/observable.js';
+
+var o1 = new AsyncGeneratorFunction(function* () {
+  // TODO(mnieper) As soon as proper async generators are implemented, they
```
arv: TODO(mnieper):

(with a colon)

test/feature/AsyncGenerators/ForOn.module.js
```diff
@@ -0,0 +1,76 @@
...
+  // should be used here.
+  yield 1;
+  yield 2;
+  yield 3;
+  return 4;
+});
+
+(async function () {
+
+  /* test for on */
```
arv: Use // comments

test/feature/AsyncGenerators/ForOn.module.js
```diff
@@ -0,0 +1,76 @@
...
+      }
+      almostEmpty.push(i);
+    }
+  } while (false);
+  assert.deepEqual(almostEmpty, [1]);
+
+  /* test return */
+  var value = await (async function () {
+    for (var i on o1) {
+      if (i == 1) {
```
arv: ===

test/feature/AsyncGenerators/ForOn.module.js
```diff
@@ -0,0 +1,76 @@
...
+      break;
+    }
+    cubes.push(i * i * i);
+  }
+  assert.deepEqual(cubes, [1, 8]);
+
+  /* test continue */
+  var list = [];
+  for (var i on o1) {
+    if (i === 2) {
```
arv: Use `===`

src/codegeneration/InnerForOnTransformer.js
mnieper: Oh, yeah. By the way, the same mistake is in FnExtractCompletions.js. I'll correct it there as well.

test/feature/AsyncGenerators/ForOn.module.js
arv: This one... `===`

test/feature/AsyncGenerators/resources/observable.js
```diff
@@ -0,0 +1,71 @@
+const generator_ = Symbol();
```
arv: no underscores for variables

test/feature/AsyncGenerators/resources/observable.js
```diff
@@ -0,0 +1,71 @@
...
+class DecoratedGenerator {
+  constructor(generator, onDone) {
+    this[generator_] = generator;
+    this[onDone_] = onDone;
+  }
+
+  next(value) {
+    var result = this[generator_].next();
+    if (result.done) {
+      this[onDone_]();
```
arv: What should the receiver (this) be here?

test/feature/Yield/ReturnCatch.js
```diff
@@ -0,0 +1,16 @@
+function* f () {
```
arv: This and next test are part of the other PR I assume?

src/codegeneration/InnerForOnTransformer.js
```diff
@@ -0,0 +1,230 @@
...
+      // {var,let} initializer = $value;
+      assignment = createVariableStatement(
+          tree.initializer.declarationType,
+          tree.initializer.declarations[0].lvalue, value);
+    } else {
+      assignment = parseStatement `
+          ${tree.initializer} = ${value};`;
+    }
+
+    var body = new Block(tree.body.location, [assignment].concat(tree.body.statements));
```
arv: or `[assignment, ...tree.body.statements]`

test/feature/AsyncGenerators/resources/observable.js
mnieper: Even in this case? Before symbols, I would write: `this.generator_`. Now I would have written: `this[generator_]` to make the property truely private.

If there was an accessor by the same name, I would get a naming problem if the symbol of the private property was also named without an underscore, wouldn't I?

arv:
A few more nits.

I'm getting excited now that I see how this is solidifying.

test/feature/Yield/ReturnCatch.js
mnieper: Was from private merge because I needed the other patch in development.

test/feature/AsyncGenerators/resources/observable.js
mnieper: You mean onDone.call(?, ...)? I don't know of a good value for ? here. Could be this, the instance of DecoratedGenerator. But in any case, this is only part of a test, and that part should later be replaced by a proper async generator.

mnieper:
I'm glad that you like it... it was a piece of work, in particular to understand how the compiler works.

arv:
The two hardest parts of the compiler are CPS transformer and the block binding transformer

mnieper:
@arv I know why compiler writers like Scheme. It's much more easy to reason about and transform code written in Scheme than code written in JavaScript.

http://matt.might.net/articles/cps-conversion/ :-)

mnieper:
Improved the code per your suggestions. (Although I think one really needs a convention for naming private symbols.)

arv:
I don't see the new code. Did you push?

mnieper:
Let me check...

src/codegeneration/InnerForOnTransformer.js
arv: This was the one that thought you had not updated the code.

arv:
LGTM. I'll fix the two remaining nits.

src/codegeneration/InnerForOnTransformer.js
mnieper: Funny, I corrected it in my editor but the edit must have been lost.

2015-02-06 18:44 GMT+01:00 Erik Arvidsson <notifications@github.com>:

PR #1693, opened 2015-02-04
Asserts that import imports only what was asked

yokomizor:
I could not found any test like this.

Maybe ```/test/features/Modules``` would be a better place, but there is no multiple imports test there.

Do you think this helpful?


arv:
This seems reasonable. Not high value but it doesn't hurt. We should do the same for commonjs and es6 modules too then.

yokomizor:
Done.

arv:
Committed as 72f83aaa22387bd1fab5e6153dcfdf10181d8495

PR #1704, opened 2015-02-06
Implement let and const in generators.

### LINKED ISSUE #1379: Traceur may produce `yield` in generator output in conjunction with block-level bindings

mnieper:
At the moment, the block binding transformer wraps loop bodies into functions. This is a problem if loop bodies do contain `yield` expressions.

Following an idea of @arv, this patch turns those generated functions into generator functions if needed and lets the outer code yield from them.

Fixes #1379.

### DIFF
--- a/src/semantics/ScopeChainBuilder.js
+++ b/src/semantics/ScopeChainBuilder.js
@@ -62,6 +62,7 @@ export class ScopeChainBuilder extends ScopeVisitor {
       this.visitAny(tree.name);
     }
     this.visitAny(tree.parameterList);
+    scope.inGenerator = tree.isGenerator();
     this.visitAny(tree.body);
     this.popScope(scope);
   }
--- a/src/codegeneration/BlockBindingTransformer.js
+++ b/src/codegeneration/BlockBindingTransformer.js
@@ -141,6 +141,11 @@ import {prependStatements} from './PrependStatements.js';
  *
  * The block binding rewrite pass assumes that deconstructing assignments
  * and variable declarations have already been desugared. See getVariableName_.
+ *
+ * Note:
+ *
+ * If the transformation happens inside a generator, the inner function
+ * becomes an inner generator.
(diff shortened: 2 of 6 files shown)


src/semantics/ScopeChainBuilder.js
```diff
@@ -62,6 +62,7 @@ export class ScopeChainBuilder extends ScopeVisitor {
       this.visitAny(tree.name);
     }
     this.visitAny(tree.parameterList);
+    scope.inGenerator = tree.isGenerator();
```
arv: Don't we need the same for FunctionDeclaration and MethodPropertyAssignment?

arv:
That turned out simpler than I expected.

src/semantics/ScopeChainBuilder.js
mnieper: There is another check in `visitFunctionBodyForScope` in `ScopeVisitor.js`.

arv:
LGTM

PR #1727, opened 2015-02-12
Implement proper tail calls.

mnieper:
The following PR implements proper tail calls as demanded by ES6. The are behind the new experimental flag `--proper-tail-calls`. (Experimental because it necessarily reduces performance of the generated code.)

This probably makes Traceur the first transpiler properly supporting tail calls!

Each function that contains calls in tail position is rewritten. All tests pass with proper tail calls enabled by default. Traceur itself compiled with `--proper-tail-calls true` also works as expected.

At the moment, the transformer does not touch arrow functions, class declarations and class expressions (so they should be transpiled if their are supposed to be properly tail recursive). Likewise, template literals in tail position are currently not touched, so the corresponding transpilation should be enabled as well.

The changes to `runtime.js` are far less worse than what the diff shows. I had to wrap most of that code into an extra IIFE and the additional two spaces at the beginning of each line cause the diffs.

It is left as a TODO to check all other transformers that they do not convert tail calls in non tail calls.

Please review this PR; it's hopefully ready to merge.

### DIFF
--- a/src/codegeneration/FromOptionsTransformer.js
+++ b/src/codegeneration/FromOptionsTransformer.js
@@ -41,6 +41,7 @@ import {ObjectLiteralTransformer} from './ObjectLiteralTransformer.js';
 import {PropertyNameShorthandTransformer} from
     './PropertyNameShorthandTransformer.js';
 import {InstantiateModuleTransformer} from './InstantiateModuleTransformer.js';
+import {ProperTailCallTransformer} from './ProperTailCallTransformer.js';
 import {RegularExpressionTransformer} from './RegularExpressionTransformer.js';
 import {RestParameterTransformer} from './RestParameterTransformer.js';
 import {SpreadTransformer} from './SpreadTransformer.js';
@@ -158,7 +159,8 @@ export class FromOptionsTransformer extends MultiTransformer {
       append(ClassTransformer);
 
     if (transformOptions.propertyMethods ||
-              transformOptions.computedPropertyNames) {
+        transformOptions.computedPropertyNames ||
+        transformOptions.properTailCalls) {
       append(ObjectLiteralTransformer);
     }
 
@@ -217,5 +219,9 @@ export class FromOptionsTransformer extends MultiTransformer {
 
     if (transformOptions.symbols)
       append(SymbolTransformer);
+
+    if (transformOptions.properTailCalls) {
+      append(ProperTailCallTransformer);
+    }
   }
 }
--- a/src/codegeneration/ObjectLiteralTransformer.js
+++ b/src/codegeneration/ObjectLiteralTransformer.js
@@ -45,9 +45,9 @@ import {propName} from '../staticsemantics/PropName.js';
  * computed property name, an at name or a __proto__ property.
  */
 class FindAdvancedProperty extends FindVisitor {
-  constructor(options) {
+  constructor(transformOptions) {
     super(true);
-    this.options_ = options;
+    this.transformOptions_ = transformOptions;
     this.protoExpression = null;
   }
 
@@ -65,17 +65,29 @@ class FindAdvancedProperty extends FindVisitor {
   }
 
   visitGetAccessor(tree) {
-    this.visitAny(tree.name);
-    // We do not want to visit object literals in the accessor's body.
+    if (this.transformOptions_.properTailCalls) {
+      // TODO(mnieper): This could first check whether there are any calls in
+      // tail position in the method's body.
+      this.found = true;
+    } else {
+      this.visitAny(tree.name);
+      // We do not want to visit object literals in the accessor's body.
+    }
   }
 
   visitSetAccessor(tree) {
-     this.visitAny(tree.name);
-    // We do not want to visit object literals in the accessor's body.
+    if (this.transformOptions_.properTailCalls) {
+      // TODO(mnieper): This could first check whether there are any calls in
+      // tail position in the method's body.
+      this.found = true;
+    } else {
+      this.visitAny(tree.name);
+      // We do not want to visit object literals in the accessor's body.
+    }
   }
 
   visitComputedPropertyName(tree) {
-    if (this.options_.transformOptions.computedPropertyNames)
+    if (this.transformOptions_.computedPropertyNames)
       this.found = true;
   }
 }
@@ -98,7 +110,7 @@ export class ObjectLiteralTransformer extends TempVarTransformer {
    */
   constructor(identifierGenerator, reporter, options) {
     super(identifierGenerator);
-    this.options_ = options;
+    this.transformOptions_ = options.transformOptions;
     this.protoExpression = null;
     this.needsAdvancedTransform = false;
     this.seenAccessors = null;
@@ -200,7 +212,7 @@ export class ObjectLiteralTransformer extends TempVarTransformer {
     var oldSeenAccessors = this.seenAccessors;
 
     try {
-      var finder = new FindAdvancedProperty(this.options_);
+      var finder = new FindAdvancedProperty(this.transformOptions_);
       finder.visitAny(tree);
       if (!finder.found) {
         this.needsAdvancedTransform = false;
--- a/src/codegeneration/ProperTailCallTransformer.js
+++ b/src/codegeneration/ProperTailCallTransformer.js
@@ -1,0 +1,133 @@
+// Copyright 2015 Traceur Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the 'License');
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an 'AS IS' BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {TempVarTransformer} from './TempVarTransformer.js';
+import {RewriteTailCallsTransformer} from './RewriteTailCallsTransformer.js';
+import {
+  createFunctionBody,
+  createFunctionExpression,
+  createIdentifierExpression as id
+} from './ParseTreeFactory.js';
+import {
+  parseExpression,
+  parseStatement,
+  parseStatements
+} from './PlaceholderParser.js';
+import {
+  AnonBlock,
+  FunctionDeclaration,
+  FunctionExpression,
+} from '../syntax/trees/ParseTrees.js';
+
+//
+// Example:
+//
+// function f(a) {
+//   var b = g(a);
+//   f;
+//   return h(b);
+// }
+//
+// Becomes:
+//
+// $traceurRuntime.initTailRecursiveFunction(function f(a) {
+//   return $traceurRuntime.call(function(a) {
+//     var b = g(a);
+//     f;
+//     return $traceurRuntime.continuation(h, null, [b]);
+//   }, this, arguments);
+// })
+
+export class ProperTailCallTransformer extends TempVarTransformer {
+  // TODO(mnieper): This transformer currently expects that classes and template
+  // literals have already been desugared. Otherwise they are not guaranteed
+  // to have proper tail calls.
+
+  constructor(identifierGenerator) {
+    super(identifierGenerator);
+    this.inBlock_ = false;
+  }
+
+  transformFunctionDeclaration(tree) {
+    var tree = super.transformFunctionDeclaration(tree);
+    if (tree.functionKind !== null) {
+      // do not transform async/generator functions
+      return tree;
+    }
+
+    var nameIdExpression = id(tree.name.identifierToken);
+
+    var setupFlagExpression = parseExpression
+        `$traceurRuntime.initTailRecursiveFunction(${nameIdExpression})`;
+
+    var funcDecl = this.transformFunction_(tree, FunctionDeclaration);
+    if (funcDecl === tree) {
+      return tree;
+    }
+
+    // Function declarations in blocks do not hoist. In that case we add the
+    // variable declaration after the function declaration.
+
+    var tmpVar = id(this.inBlock_ ?
+        this.getTempIdentifier() : this.addTempVar(setupFlagExpression));
+
+    if (!this.inBlock_) {
+      return funcDecl;
+    }
+
+    return new AnonBlock(null, [
+      funcDecl,
+      parseStatement `var ${tmpVar} = ${setupFlagExpression};`
+    ]);
+  }
+
+  transformFunctionExpression(tree) {
+    var tree = super.transformFunctionExpression(tree);
+    if (tree.functionKind) {
+      // do not transform async/generator functions
+      return tree;
+    }
+
+    var functionExpression =
+        this.transformFunction_(tree, FunctionExpression);
+
+    if (functionExpression === tree) {
+      return tree;
+    }
+
+    return parseExpression `
+        $traceurRuntime.initTailRecursiveFunction(${functionExpression})`;
+  }
+
+  transformFunction_(tree, constructor) {
+    var body = RewriteTailCallsTransformer.transform(this, tree.body);
+    if (body === tree.body) {
+      return tree;
+    }
+    var func = id(this.getTempIdentifier());
+    var innerFunction = createFunctionExpression(tree.parameterList, body);
+    var outerBody = createFunctionBody(parseStatements `
+        return $traceurRuntime.call(${innerFunction}, this, arguments);`);
+    return new constructor(tree.location, tree.name, tree.functionKind,
+        tree.parameterList, tree.typeAnnotation, tree.annotations, outerBody);
+  }
+
+  transformBlock(tree) {
+    var inBlock = this.inBlock_;
+    this.inBlock_ = true;
+    var rv = super.transformBlock(tree);
+    this.inBlock_ = inBlock;
+    return rv;
+  }
+}
--- a/src/codegeneration/RewriteTailExpressionsTransformer.js
+++ b/src/codegeneration/RewriteTailExpressionsTransformer.js
@@ -1,0 +1,165 @@
+// Copyright 2015 Traceur Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the 'License');
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an 'AS IS' BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {ParseTreeTransformer} from './ParseTreeTransformer.js';
+import {
+  ArgumentList,
+  BinaryExpression,
+  CallExpression,
+  ConditionalExpression,
+  MemberExpression,
+  MemberLookupExpression
+} from '../syntax/trees/ParseTrees.js';
+import {
+  createArrayLiteralExpression,
+  createAssignmentExpression,
+  createCommaExpression,
+  createMemberExpression,
+  createIdentifierExpression as id,
+  createNullLiteral,
+  createParenExpression,
+} from './ParseTreeFactory.js';
+import {
+  COMMA_EXPRESSION,
+  MEMBER_EXPRESSION,
+  MEMBER_LOOKUP_EXPRESSION,
+  IDENTIFIER_EXPRESSION,
+  PAREN_EXPRESSION,
+  THIS_EXPRESSION
+} from '../syntax/trees/ParseTreeType.js';
+import {
+  AND,
+  OR
+} from '../syntax/TokenType.js';
+
+function createCall(tree, operand, thisArg) {
+  var argList = tree.args; // can be null
+  var argArray = argList ? argList.args : [];
+  argArray = argArray.map(arg => {
+    if (arg.type === COMMA_EXPRESSION) {
+      return createParenExpression(arg.type);
+    }
+    return arg;
+  });
+  return new CallExpression(tree.location,
+      createMemberExpression('$traceurRuntime', 'continuation'),
+      new ArgumentList(argList ? argList.location : null, [operand, thisArg,
+          createArrayLiteralExpression(argArray)]));
+}
+
+export class RewriteTailExpressionsTransformer extends ParseTreeTransformer {
+  // TODO(mnieper): template literals in tail position
+
+  constructor(bodyTransformer) {
+    this.bodyTransformer_ = bodyTransformer;
+  }
+
+  transformBinaryExpression(tree) {
+    var operator = tree.operator;
+    if (operator.type !== AND && operator.type !== OR) {
+      return tree;
+    }
+    var right = this.transformAny(tree.right);
+    if (right !== tree.right) {
+      return new BinaryExpression(tree.location, tree.left, operator, right);
+    }
+    return tree;
+  }
+
+  transformCallExpression(tree) {
+    var operand = tree.operand;
+    while (operand.type === PAREN_EXPRESSION) {
+      operand = operand.expression;
+      // TODO(mnieper): Readd parens to the resulting call.
+    }
+    switch (operand.type) {
+      case IDENTIFIER_EXPRESSION:
+        return createCall(tree, operand, createNullLiteral());
+      case MEMBER_EXPRESSION:
+      case MEMBER_LOOKUP_EXPRESSION:
+        return this.transformMemberExpressionCall_(tree, operand)
+    }
+    return tree;
+  }
+
+  transformMemberExpressionCall_(tree, operand) {
+    var object = operand.operand;
+    var thisArg;
+    var assignment;
+    if (object.type === IDENTIFIER_EXPRESSION ||
+        object.type === THIS_EXPRESSION) {
+        // If we wanted to be strict, we would have to leave out identifier
+        // expressions here. When referring to getters on the global object,
+        // they could have side effects.
+      thisArg = object;
+    } else {
+      thisArg = id(this.bodyTransformer_.addTempVar());
+      assignment = createAssignmentExpression(thisArg, operand.operand);
+    }
+    if (operand.type === MEMBER_EXPRESSION) {
+      operand = new MemberExpression(operand.location, thisArg, operand.memberName);
+    } else {
+      operand = new MemberLookupExpression(operand.location, thisArg, operand.memberExpression);
+    }
+    if (assignment) {
+      return createParenExpression(createCommaExpression([assignment,
+          createCall(tree, operand, thisArg)]));
+    } else {
+      return createCall(tree, operand, thisArg);
+    }
+  }
+
+  transformCommaExpression(tree) {
+    var expressions = tree.expressions;
+    var expression = expressions[expressions.length - 1];
+    var transformedExpression = this.transformAny(expression);
+    if (expression !== transformedExpression) {
+      expressions = expressions.slice(0, -1);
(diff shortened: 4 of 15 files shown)


src/codegeneration/FromOptionsTransformer.js
```diff
@@ -158,7 +159,8 @@ export class FromOptionsTransformer extends MultiTransformer {
       append(ClassTransformer);
 
     if (transformOptions.propertyMethods ||
-              transformOptions.computedPropertyNames) {
+              transformOptions.computedPropertyNames ||
```
arv: wrong indentation

src/codegeneration/ProperTailCallTransformer.js
```diff
@@ -0,0 +1,137 @@
...
+ *   var b = g(a);
+ *   f;
+ *   return h(b);
+ * }
+ *
+ * Becomes:
+ *
+ * $traceurRuntime.initTailRecursiveFunction(function f(a) {
+ *   return $traceurRuntime.call(function(a) {
+ *     b = g(a);
```
arv: `var b` I assume?

src/codegeneration/ProperTailCallTransformer.js
```diff
@@ -0,0 +1,137 @@
...
+        $traceurRuntime.initTailRecursiveFunction(${functionExpression})`;
+  }
+
+  transformFunction_(tree, constructor) {
+    var body = RewriteTailCallsTransformer.transform(this, tree.body);
+    if (body === tree.body) {
+      return tree;
+    }
+    var func = id(this.getTempIdentifier());
+    var parameterList = CloneTreeTransformer.cloneTree(tree.parameterList);
```
arv: Maybe add a comment explaining why this work around is wanted?

src/codegeneration/ProperTailCallTransformer.js
```diff
@@ -0,0 +1,137 @@
...
+    var innerFunction = createFunctionExpression(parameterList, body);
+    var outerBody = createFunctionBody(parseStatements `
+        return $traceurRuntime.call(${innerFunction}, this, arguments);`);
```
arv: We might want to consider alpha renaming this and arguments instead?

src/codegeneration/ProperTailCallTransformer.js
```diff
@@ -0,0 +1,137 @@
...
+    return new constructor(tree.location, tree.name, tree.functionKind,
+        tree.parameterList, tree.typeAnnotation || null,
```
arv: This `|| null` should not be required.

src/codegeneration/RewriteTailExpressionsTransformer.js
```diff
@@ -0,0 +1,163 @@
...
+  IDENTIFIER_EXPRESSION,
+  PAREN_EXPRESSION,
+  THIS_EXPRESSION
+} from '../syntax/trees/ParseTreeType.js';
+import {
+  AND,
+  OR
+} from '../syntax/TokenType.js';
+
+function createCall(tree, operand, thisArg) {
```
arv: Maybe this belongs in ParseTreeFactory.js

arv: never mind

src/codegeneration/RewriteTailExpressionsTransformer.js
```diff
@@ -0,0 +1,163 @@
...
+    return tree;
+  }
+
+  transformMemberExpressionCall_(tree, operand) {
+    var object = operand.operand;
+    var thisArg;
+    var assignment;
+    if (object.type === IDENTIFIER_EXPRESSION ||
+        object.type === THIS_EXPRESSION) {
+        // TODO(mnieper): More checks. Is there a better way?
```
arv: What is your intent here?

arv: I see.

Strictly speaking even identifier expression might have side effects but we haven't cared about that in the past.

src/codegeneration/RewriteTailExpressionsTransformer.js
```diff
@@ -0,0 +1,163 @@
...
+  transformArrayLiteralExpression(tree) {return tree;}
+  transformArrowFunctionExpression(tree) {return tree;}
+  transformFunctionExpression(tree) {return tree;}
+  transformIdentifierExpression(tree) {return tree;}
+  transformLiteralExpression(tree) {return tree;}
+  transformMemberExpression(tree) {return tree;}
+  transformMemberLookupExpression(tree) {return tree;}
+  transformPostfixExpression(tree) {return tree;}
+  transformObjectLiteralExpression(tree) {return tree;}
+  transformThisExpression(tree) {return tree;}
```
arv: this one seems unnecessary.

src/runtime/runtime.js
```diff
@@ -96,291 +83,418 @@
...
-    symbolValues[key] = this;
+  function isTailRecursive(func) {
+    return func[isTailRecursiveName] || false;
   }
-  $defineProperty(SymbolValue.prototype, 'constructor', nonEnum(Symbol));
-  $defineProperty(SymbolValue.prototype, 'toString', {
-    value: Symbol.prototype.toString,
-    enumerable: false
+
+  function tailCall(func, thisArg, argArray) {
```
arv: It would be nice if this could be moved to its own file.

arv:
I was thinking a bit more how to better tests this instead of just hoping we would run out of stack. Maybe we could write a non feature test that inspects the continuation somehow. Another option might be to throw an error and look at the error .stack.

src/codegeneration/ProperTailCallTransformer.js
mnieper: I need a verbatim copy of the parameter list because it is used in function bodies. In case there are subsequent transformers, they need to be able to transform the two parameter lists individually.

src/codegeneration/ProperTailCallTransformer.js
mnieper: This won't work because the inner function is not always called (by $traceurRuntime.call) with the same parameters as the wrapping function was called - nor do I see a need for this.

src/codegeneration/ProperTailCallTransformer.js
mnieper: I took that from the GeneratorTransformPass, so I thought this was necessary.

src/codegeneration/RewriteTailExpressionsTransformer.js
mnieper: Interesting... what could be the side effect of an identifier expression?

src/codegeneration/RewriteTailExpressionsTransformer.js
mnieper: Is a ThisExpression subsumed by a LiteralExpression?

src/runtime/runtime.js
mnieper: I had had this in an extra module `tailCalls.js` until I wanted to compile the whole Traceur codebase with PTC support. Then I ran into the issue that the transformed runtime itself needed bits of `tailCalls.js`, so I had to move it at the top. Because `tailCalls.js` needed `$traceurRuntime` to add properties, I merged it into the main runtime file.

A probably better solution (which would add more globals, though) would be to change `$traceurRuntime.tailcall` to `$traceurRuntime$tailCall`, etc., so that the PTC part of the runtime can be loaded at the very beginning independent of `$traceurRuntime` already set up.

As every second call in a PTC-transpiled program will be a call to `$traceurRuntime.call`, it may even speed things up if one removes the extra property lookup on `$traceurRuntime` by going global.

What do you think?

mnieper:
I'll see how I can possibly reformulate the tests. --- I have added quite a number of comments above, please review them.

mnieper:
This is one idea how we can measure the stack size:

```javascript
Error.stackTraceLimit = Infinity;

var stackSize;
function recordStackSize() {
  try {
    Error.prepareStackTrace = function(error, structuredStackTrace) {
      return structuredStackTrace.length;
    }
    stackSize = new Error().stack;
  } finally {
    delete Error.prepareStackTrace;
  }
}

function assertProperTailCalls(limit = 30) {
  assert.isTrue(stackSize < limit);
}
```

I have checked that the test environment accounts for about 15 stack frames, so the limit of 30 is very safe when one has only about 40 or 50 tail calls and 100,000.
 
Could this go in one file of the testing framework in test?

src/codegeneration/ProperTailCallTransformer.js
arv: That should be fine. Trees are immutable so if one branch needs to be transformed it will return a new tree.

src/codegeneration/ProperTailCallTransformer.js
arv: OK.

src/codegeneration/ProperTailCallTransformer.js
arv: There might be a bug somewhere else that caused the original code to use this pattern. In Traceur we are very strict with making sure we set all properties to null and do not leave them undefined.

src/codegeneration/RewriteTailExpressionsTransformer.js
arv: The global object might have getters that have side effects.

src/codegeneration/RewriteTailExpressionsTransformer.js
arv: No, but transformThisExpression just returns the tree.

src/runtime/runtime.js
arv: One of my motivations here was that we could make this part configurable. Not many people will use PTC so it would be good if they could skip this part of the runtime library.

Changing `$traceurRuntime.foo` to `$traceurRuntime$foo` is something I don't want to do at this point. We have a lot of work to do with our runtime and I'm not sure whether that will make things better or worse.

src/codegeneration/ProperTailCallTransformer.js
mnieper: Okay, will check whether it works without `|| null`. I have recently run into the issue that I got an undefined `functionKind` in the parsetree.

src/codegeneration/RewriteTailExpressionsTransformer.js
mnieper: Ah, of course. Is there an easy check to see whether a given identifier is in scope? Or would I have to reimplement something complicated like the block binding transformer machinery?

In that case, I'd restrict here to the `this` case.

src/runtime/runtime.js
mnieper: One idea could be to let the compiler add a tiny bit of code at the top of every module or script that uses tail calls. The only problem with this approach is that I need a unique global identity `CONTINUATION_TYPE`. Any idea how to get this?

Another idea could be to allow `$traceurRuntime` to be already setup before `runtime.js` is executed and to use a known property on `$traceurRuntime` to detect whether the `runtime.js` has already been loaded.

src/codegeneration/ProperTailCallTransformer.js
arv: Probably this one: 9ec2b9f043447aec6321226c8ed26af2b50a545a

src/codegeneration/RewriteTailExpressionsTransformer.js
arv: We have a lot of cases where we assume an IdentifierExpression has no side effect.

src/runtime/runtime.js
arv: I'm not sure I understand the `CONTINUATION_TYPE` issue?

mnieper: I need an unforgeable unique identity for two things: A transpiled function can either return a value as always or a continuation (consisting of another function and its args). The function `$traceurRuntime.call` need to have a way to distinguish these types.

Secondly, I need to call a transpiled function in two ways: It has to act normally in case of a caller who doesn't know about the transformation, and it is allowed to return continuation records in case it is called in a special way. It also uses the unique unforgeable identity to discern these two cases.

mnieper: A new thought:

The amount of code I added to `runtime.js` produces only a very tiny amount of overhead so this should not be an issue for most use cases where PTCs are turned off.

What does make a difference, though, is the redefinition of `Function.prototype.apply` and `Function.prototype.call`. But I can defer changing their implementation until the first call to `initTailRecursiveFunction`, which will never happen in code that is not compiled with `--proper-tail-calls true`.

arv: I'm also concerned about the binary size of the runtime. When deploying to the web I doubt we want to enable PTC and therefore we should not include the runtime library support for it.

Still, I don't think this should block the work you are doing. We needed to solve this issue even before this PR.

arv: Can the runtime provide the sentinel?

mnieper: You mean the runtime just sets up the sentinel and leaves the rest of the code to the per module initialization code? This would work if we didn't have the chicken-and-egg problem when the runtime itself is transpiled with PTCs.

Okay, so I will leave the new code in the runtime but disable modifying apply/call when not needed. And then we can think of a more global solution as you suggest.

arv: Sounds good to me

mnieper:
Issues have been taken care of in the new commit.

Tests run much faster now :-)

Although the 100,000 was quite impressive...

P.S.: Who is responsible for updating the table by kangax?

arv:
That would be @kangax ;-)

src/codegeneration/ObjectLiteralTransformer.js
```diff
@@ -64,13 +64,25 @@ class FindAdvancedProperty extends FindVisitor {
   }
 
   visitGetAccessor(tree) {
-    this.visitAny(tree.name);
-    // We do not want to visit object literals in the accessor's body.
+    if (this.options_.transformOptions.properTailCalls) {
```
arv: Maybe add 

```js
this.transformOptions_ = options.transformOptions;
```

to the constructor?

src/codegeneration/ProperTailCallTransformer.js
```diff
@@ -0,0 +1,134 @@
...
+  parseStatement,
+  parseStatements
+} from './PlaceholderParser.js';
+import {
+  AnonBlock,
+  FunctionDeclaration,
+  FunctionExpression,
+} from '../syntax/trees/ParseTrees.js';
+
+/*
```
arv: Use // for comments. Use /** for JSDoc.

src/codegeneration/RewriteTailExpressionsTransformer.js
```diff
@@ -0,0 +1,164 @@
...
+    var right = this.transformAny(tree.right);
+    if (right !== tree.right) {
+      return new BinaryExpression(tree.location, tree.left, operator, right);
+    }
+    return tree;
+  }
+
+  transformCallExpression(tree) {
+    var operand = tree.operand;
+    while (operand.type === PAREN_EXPRESSION) {
```
arv: We add parens in a few places to make the stack trace better. Maybe we should readd them after the transform? I don't think it is worth doing since our stack traces will already be hard to decipher.

src/codegeneration/RewriteTailExpressionsTransformer.js
```diff
@@ -0,0 +1,164 @@
...
+    return tree;
+  }
+
+  transformMemberExpressionCall_(tree, operand) {
+    var object = operand.operand;
+    var thisArg;
+    var assignment;
+    if (object.type === IDENTIFIER_EXPRESSION ||
+        object.type === THIS_EXPRESSION) {
+        // XXX If we wanted to be strict, we would have to leave out identifier
```
arv: I don't think we use XXX anywhere in our comments. Just remove that. The comment reads fine without it.

src/codegeneration/RewriteTailExpressionsTransformer.js
```diff
@@ -0,0 +1,164 @@
...
+      assignment = createAssignmentExpression(thisArg, operand.operand);
+    }
+    if (operand.type === MEMBER_EXPRESSION) {
+      operand = new MemberExpression(operand.location, thisArg, operand.memberName);
+    } else {
+      operand = new MemberLookupExpression(operand.location, thisArg, operand.memberExpression);
+    }
+    if (assignment) {
+      return createParenExpression(createCommaExpression([assignment,
+        createCall(tree, operand, thisArg)]));
```
arv: +2 indent

test/feature/ProperTailCalls/Call.js
```diff
@@ -0,0 +1,37 @@
+// Options: --proper-tail-calls
+
+Error.stackTraceLimit = Infinity;
```
arv: Maybe move to test-utils.js?

kangax:
@mnieper please send PR on https://github.com/kangax/compat-table and I'll gladly merge it in! Thanks.

src/runtime/runtime.js
```diff
@@ -25,8 +25,8 @@
 
   var $Object = Object;
   var $TypeError = TypeError;
-  var $create = $Object.create;
   var $defineProperties = $Object.defineProperties;
+  var $create = $Object.create;
```
arv: no need for this change

src/runtime/runtime.js
```diff
@@ -37,29 +37,27 @@
...
+    var argArray = [thisArg];
+    for (var i = 0; i < args.length; ++i) {
+      argArray[i + 1] = args[i];
+    }
+    var func = $apply(Function.prototype.bind, operand, argArray);
+    return func;
   }
 
-  var method = nonEnum;
+  function $construct(func, argArray) {
```
arv: The idea behind the leading $ was to use that for the original version of the built in. Since there is no builtin construct (ES6 has `Reflect.construct` but that is not relevant here).

src/runtime/runtime.js
```diff
@@ -37,29 +37,27 @@
...
-    return {
-      configurable: true,
-      enumerable: false,
-      value: value,
-      writable: true
-    };
+  function $bind(operand, thisArg, args) {
+    // args may be an arguments-like object
+    var argArray = [thisArg];
+    for (var i = 0; i < args.length; ++i) {
```
arv: `i++` for consistency

src/runtime/runtime.js
```diff
@@ -96,291 +83,427 @@
...
+    // Function.prototype.apply do proper tail calls.
+
+    Function.prototype.call = initTailRecursiveFunction(
+        function call(thisArg) {
+          var result = tailCall(function (thisArg) {
+            var argArray = [];
+            for (var i = 1; i < arguments.length; ++i) {
+              argArray[i - 1] = arguments[i];
+            }
+            var continuation = createContinuation(this, thisArg, argArray);
```
arv: skip continuation var here

arv: Oh... maybe this is to prevent the tail call?

Add a comment if so.

src/runtime/runtime.js
```diff
@@ -96,291 +83,427 @@
...
-
-  function SymbolValue(description) {
-    var key = newUniqueString();
-    $defineProperty(this, symbolDataProperty, {value: this});
-    $defineProperty(this, symbolInternalProperty, {value: key});
-    $defineProperty(this, symbolDescriptionProperty, {value: description});
-    freeze(this);
-    symbolValues[key] = this;
+  function isTailRecursive(func) {
+    return func[isTailRecursiveName] || false;
```
arv:
```js
return !!func[isTailRecursiveName];
```

but in this case it is not important to return a boolean.

src/runtime/runtime.js
```diff
@@ -96,291 +83,427 @@
...
-      value: undefined
-    },
-    self: {
-      value: undefined
+
+  function tailCall(func, thisArg, argArray) {
+    var continuation = argArray[0];
+    if (isContinuation(continuation)) {
+      continuation = $apply(func, thisArg, continuation[3]);
+      return continuation;
```
arv: Can you add a comment to every place in this file where you do not want the PTC to be applied?

src/runtime/runtime.js
mnieper: I have `construct` below as well. I coined this one `$construct` because of its relation to `Reflect.construct` and for similarity to `$apply` and `Prototype.function.apply`, etc.

src/runtime/runtime.js
mnieper: Yep.

mnieper: There is a global comment above that piece in the runtime.

src/runtime/runtime.js
arv: Ok. Fair enough.

src/codegeneration/RewriteTailExpressionsTransformer.js
mnieper: I'll add a TODO

test/feature/ProperTailCalls/Call.js
mnieper: I'll leave this to another PR if you agree.

src/runtime/runtime.js
mnieper: Wasn't intended.

mnieper:
Pushed the corrections.

mnieper:
@kangax Thanks. I forked your project and took a look at it. But I'd like someone else to run the tests because the only browsers I have running are the latest Firefox and the latest Chromium, so this would probably give false results.

arv:
LGTM

arv:
Committed as d9cc38aa813a9591bb72e8fd8545984d45919498

mnieper:
Cool.

arv:
Yes. Extremely cool.

mnieper:
In the end it was easier than I thought. :-)

kangax:
@mnieper I guess I can just change the result of TCO test from `false` to `true` for Traceur and that's it, right?

mnieper:
@kangax Both TCO tests should pass. (There are equivalent tests in Traceur's test suite.)

Apart from that, %GeneratorPrototype%.return also does work now in Traceur.

For ES7, async generators have been implemented. (https://github.com/jhusain/asyncgenerator)

Finally, could you please review the test here: https://github.com/google/traceur-compiler/issues/1730. I think most transpilers are broken in this regard and this could be a test worth adding to your table.

kangax:
@mnieper thanks, I updated Traceur's ES6 results. For ES7, I don't believe we have a test for async generators (only async functions and async arrow functions). I'll get to that at some point. I'll also take a look at #1730 later if I get a chance — no time at the moment!

kangax:
@sebmck looks like someone needs to step up their game ;)

mnieper:
@sebmck I have already been asked whether I could implement this work and my work on async generators in 6to5. However, as I am doing everything in my spare time, I'd rather like to do something new than just reimplement something (and I'd have to go through another codebase to understand it).

Nevertheless, I hope you can take a look at this stuff and maybe find another volunteer. In any case, I will readily answer any question.

sebmck:
@kangax I'm the only one working on the 6to5 internals and I do it all in my spare time so I really wish I could "step up my game". Also don't mean to be pedantic but the Traceur TCO tests should be marked as "flag" on compat-table since you have to explicitly enable the `properTailCalls` option.

@mnieper Awesome work! Will definently look into it when I get the time. What's your preferred way to be contacted? Would love to talk about it.

TCO was disabled because there were issues with nested tail calls returning their tail call instance. TRO is currently enabled by default though and actually makes recursive calls **signifcantly** faster:

```javascript
function fact(n) {
  function f(n, acc) { return  n === 0 ? acc : f(n - 1, acc * n); }
  return f(n, 1);
}
fact(20);
```

[jsperf screenshot](http://jsperf.com/6to5-factorial-tco/2)

Might be worth looking into this to see if it can be implemented in Traceur by default. The main issue we ran into with TCO was it significantly slowed down a lot of calls that didn't need to be, what's the performance impact like?

mnieper:
@sebmck 

*I'm the only one working on the 6to5 internals and I do it all in my spare time*

Wow!

I haven't implemented special optimisations for self-calling functions - in fact, the code hasn't been optimised in any way yet. Running all tests with proper tail calls turned on takes 10 times the normal time (and from the way it has to be implemented in general, one has to expect a factor in that order).

For easy cases like function calling itself, yes, one should add an extra transformation pass to eliminate the recursion altogether. But as JavaScript has a lot of imperative features to write loops, I don't think this will be the majority of use cases for PTCs. I think they are much more important for programming in continuation passing style, combinator libraries and JavaScript as a compilation target for languages relying on PTCs. In this case, however, just optimising self-calling recursive functions won't help much.

For PTC to be really useful, we will wait for the browser engines to implement it (any news on this by the V8 and SpiderMonkey teams?). But with transpilers like Traceur and 6to5 allowing PTCs (albeit at lower performance) it allows library writers to already use this feature so that the libraries are already ready for production when PTC is implemented in the browser engines.

You can either contact me via email (marc.nieper@gmail.com) or via Google Talk/Jabber at the same address.

kangax:
@sebmck I mostly meant it as a joke — 6to5 is already doing amazing :) We'll take care of flag situation shortly; thanks for noticing!

sebmck:
@kangax Sorry! In hindsight that much was obvious, had only just woken up :stuck_out_tongue_winking_eye:

mnieper:
@kangax How do you handle the flag thing? At the moment, the more options a transpiler offers, the worse seems to be its display on your table, doesn't it?

You would also have to add flags to all ES7 language elements Traceur supports because they are also hidden behind a flag (for the benefit of the user).

sebmck:
@mnieper There's a separate table for ES7: https://kangax.github.io/compat-table/es7/

mnieper:
@sebmck Yes, but all ES7 features of Traceur are behind  flags, which isn't shown on that table.

arv:
@sebmck 6to5 is really getting good. Like 6to5, traceur is also all done in our spare time (I try to use my 20% time but it is rare that I actually can do that). @johnjbarton is in a similar position.

@mnieper I work on V8 but PTC has not gotten on our radar yet. We have a lot of other important work to do before we will get to that.

sebmck:
@arv Thanks! I always get people thinking that I have some vendetta against Traceur when really it's the opposite and I really appreciate what you guys do and I'm super happy that both projects exist. Truth is that 6to5 (now Babel) and Traceur are targeting completely different ecosystems, Traceur with AtScript/Angular while 6to5 focuses on JSX/React.

arv:
@sebmck I didn't mean to imply that. I'm sorry if it came out that way.

I'm also don't feel happy with saying that Traceur is targetting AtScript/Angular. They liked how easy Traceur is to extend so they decided to add the transformation passes they wanted.

To me Traceur was a test bed for harmony features. We implemented a lot of the ES6 features before we brought them to TC39. Today I feel like we are still struggling to make traceur a useful tool for real world projects, something it seems Babel (nee 6to5) is doing much better at.

sebmck:
@arv No, you didn't imply that at all. It's just that people always assume there's some sort of bitter rivalry between projects in the same space.

> I also don't feel happy with saying that Traceur is targetting AtScript/Angular. They liked how easy Traceur is to extend so they decided to add the transformation passes they wanted.

I wasn't trying to say that these are in anyways locked into particular frameworks. Both projects can find use anywhere, but Traceur has become most popular within the Angular community as 6to5 has become in the React community.

> Today I feel like we are still struggling to make traceur a useful tool for real world projects, something it seems Babel (nee 6to5) is doing much better at.

Thanks, it means a lot! Babel owes a great deal to Traceur and the standards it's created, it wouldn't exist without it and I'm incredibly grateful.

mnieper:
@arv @sebmck 
I think the whole developer community can consider itself fortunate that there are at least two excellent ES6 to ES5 transpilers. The more tools there are that are able to prove that some suggested features for ES7, 8, 9, etc. is actually working and will make coding in JavaScript easier, the more likely it is that this feature will be accepted in a new standard of the language.

So I really hope that you both continue your work on your tools so that both will be used a lot in the JavaScript community.

(By way of comparison, if only V8 but no SpiderMonkey, or vice versa, were in daily use, JavaScript and its engines wouldn't have seen such a rapid progress recently. asm.js is one example that readily comes to my mind.)

PR #1729, opened 2015-02-12
Avoid unnecessary object literal transformations.

### LINKED ISSUE #1728: ObjectLiteralTransformer transforms more than needed
The ObjectLiteralTransformer transforms this piece of code

```javascript
var outer = {
  inner: {
    [0]: 1
  }
};
```

into

```javascript
outer = ($__1 = {}, Object.defineProperty($__1, "inner", {
  value: ($__0 = {}, Object.defineProperty($__0, 0, {
    value: 1,
    configurable: true,
    enumerable: true,
    writable: true
  }), $__0),
  configurable: true,
  enumerable: true,
  writable: true
}), $__1);
```

While correct, the rewriting of the outer literal is unnecessary. A correction would be to not to let the `FindAdvancedProperty` FindVisitor in `ObjectLiteralTransformer` traverse the property values itself.

mnieper:
Example:

```js
var outer = {
  inner: {
    [0]: 1
  },
  method() {
    return {[2]: 3};
  },
  set a(value) {
    return {[4]: 5};
  }
};
```

The outer literal should not be transformed.

Fixes #1728.

### DIFF
--- a/src/codegeneration/ObjectLiteralTransformer.js
+++ b/src/codegeneration/ObjectLiteralTransformer.js
@@ -54,7 +54,23 @@ class FindAdvancedProperty extends FindVisitor {
     if (isProtoName(tree.name))
       this.protoExpression = tree.value;
     else
-      super.visitPropertyNameAssignment(tree);
+      this.visitAny(tree.name);
+      // We do not want to visit object literals in the property's value.
+  }
+
+  visitPropertyMethodAssignment(tree) {
+    this.visitAny(tree.name);
+    // We do not want to visit object literals in the method's body.
+  }
+
+  visitGetAccessor(tree) {
+    this.visitAny(tree.name);
+    // We do not want to visit object literals in the accessor's body.
+  }
+
+  visitSetAccessor(tree) {
+     this.visitAny(tree.name);
+    // We do not want to visit object literals in the accessor's body.
   }
 
   visitComputedPropertyName(tree) {


mnieper:
This PR is also part of #1727 so when #1727 is included, this can be closed.

src/codegeneration/ObjectLiteralTransformer.js
```diff
@@ -54,7 +54,19 @@ class FindAdvancedProperty extends FindVisitor {
     if (isProtoName(tree.name))
       this.protoExpression = tree.value;
     else
-      super.visitPropertyNameAssignment(tree);
+      return super.visitAny(tree.name);
```
arv: No return. The visit functions return void.

src/codegeneration/ObjectLiteralTransformer.js
```diff
@@ -54,7 +54,19 @@ class FindAdvancedProperty extends FindVisitor {
...
+  }
+
+  visitPropertyMethodAssignment(tree) {
+    return super.visitAny(tree.name);
```
arv: Just do:

```js
this.visitAny(tree.name);
```

no need to go through super here.

This could also use a comment.

src/codegeneration/ObjectLiteralTransformer.js
mnieper: In case I don't know, I usually use `return` as it always does the right thing whether the function returns void or not.

With proper tail calls it also makes a semantic difference; with return the stack does not grow (not that it matters here, but I think this makes it the better idiom).

mnieper:
Pushed the changes you asked for.

arv:
Committed as f6d7bda27e05c3e14e0987b3cbf0531e877630c0

PR #1781, opened 2015-03-01
Use function state for generator

arv:
This one comes after #1779 and it simplifies handling of allowYield.

Only look at the last commit to get a sense for where this is going.


arv:
@johnjbarton

Issue #1708, opened 2015-02-06
Transpile to ES6: drop field declarations but keep initializers


tbosch:
Right now when transpiling from AtScript to ES6 traceur keeps field declarations, but I think they should be dropped, as they are not part of ES6.

Input:
```
class Test {
  a:number = 10;
}
```

Output right now:
```
class Test {
  a: $traceurRuntime.type.number;
}
```

Note: Can't provide a repl example due to https://github.com/google/traceur-compiler/issues/1707

arv:
For more background see https://gist.github.com/jeffmo/054df782c05639da2adb

PR #1791, opened 2015-03-04
Member variables in ES6

vicb:
relates to #1708 

/cc @arv @caitp

TODO:
- [x] generated ctor if super class - currently ```parseStatement `super(...arguments)` ``` fails,
- [x] the coma expression (class expression) is not valid: `let` and ";",
- [x] add tests

### DIFF
--- a/src/codegeneration/ClassTransformer.js
+++ b/src/codegeneration/ClassTransformer.js
@@ -29,14 +29,11 @@ import {
   createIdentifierToken
 } from '../codegeneration/ParseTreeFactory.js';
 import {
-  CALL_EXPRESSION,
   COMPUTED_PROPERTY_NAME,
-  EXPRESSION_STATEMENT,
   GET_ACCESSOR,
   PROPERTY_METHOD_ASSIGNMENT,
   PROPERTY_VARIABLE_DECLARATION,
   SET_ACCESSOR,
-  SUPER_EXPRESSION
 } from '../syntax/trees/ParseTreeType.js';
 import {SuperTransformer} from './SuperTransformer.js';
 import {TempVarTransformer} from './TempVarTransformer.js';
@@ -65,6 +62,10 @@ import {
 } from './PlaceholderParser.js';
 import {propName} from '../staticsemantics/PropName.js';
 import {prependStatements} from './PrependStatements.js';
+import {
+  transformConstructor,
+  getInstanceInitExpression,
+} from './MemberVariableConstructorTransformer.js';
 
 // Interaction between ClassTransformer and SuperTransformer:
 // - The initial call to SuperTransformer will always be a transformBlock on
@@ -144,7 +145,6 @@ export class ClassTransformer extends TempVarTransformer {
    */
   constructor(identifierGenerator, reporter, options) {
     super(identifierGenerator);
-    this.options_ = options;
     this.strictCount_ = 0;
     this.state_ = null;
     this.reporter_ = reporter;
@@ -198,7 +198,6 @@ export class ClassTransformer extends TempVarTransformer {
     this.state_ = {hasSuper: false};
     let superClass = this.transformAny(tree.superClass);
 
-    let hasConstructor = false;
     let protoElements = [], staticElements = [];
     let initInstanceVars = [], initStaticVars = [];
     let constructor;
@@ -226,7 +225,6 @@ export class ClassTransformer extends TempVarTransformer {
 
         case PROPERTY_METHOD_ASSIGNMENT:
           if (!tree.isStatic && propName(tree) === CONSTRUCTOR) {
-            hasConstructor = true;
             constructor = tree;
           } else {
             let transformed = this.transformPropertyMethodAssignment_(
@@ -236,7 +234,7 @@ export class ClassTransformer extends TempVarTransformer {
           break;
 
         case PROPERTY_VARIABLE_DECLARATION:
-          this.transformAny(tree);
+          tree = this.transformAny(tree);
           if (tree.initializer !== null) {
             initVars.push(tree);
           }
@@ -249,15 +247,15 @@ export class ClassTransformer extends TempVarTransformer {
 
     let object = createObjectLiteralExpression(protoElements);
     let staticObject = createObjectLiteralExpression(staticElements);
-    let initStatements = getInstanceInitStatements(initInstanceVars);
+    let initExpression = getInstanceInitExpression(initInstanceVars);
     let func;
 
-    if (!hasConstructor) {
-      func = this.getDefaultConstructor_(tree, internalName, initStatements);
+    if (!constructor) {
+      func = this.getDefaultConstructor_(tree, internalName, initExpression);
     } else {
-      if (this.options_.memberVariables) {
-        constructor = this.appendInstanceInitializers_(constructor,
-            initStatements, tree.superClass);
+      if (initInstanceVars.length > 0) {
+        constructor = transformConstructor(constructor, initExpression,
+            tree.superClass);
       }
       let homeObject = createMemberExpression(internalName, 'prototype');
       let transformedCtor = this.transformPropertyMethodAssignment_(
@@ -432,49 +430,32 @@ export class ClassTransformer extends TempVarTransformer {
     return transformedTree;
   }
 
-  getDefaultConstructor_(tree, internalName, initStatements) {
+  getDefaultConstructor_(tree, internalName, initExpression) {
     let constructorParams = createEmptyParameterList();
     let constructorBody;
+    let initStatement = createExpressionStatement(initExpression);
     if (tree.superClass) {
       let statement = parseStatement `$traceurRuntime.superConstructor(
           ${internalName}).apply(this, arguments)`;
-      constructorBody = createFunctionBody([statement, ...initStatements]);
+      constructorBody = createFunctionBody([statement, initStatement]);
       this.state_.hasSuper = true;
     } else {
-      constructorBody = createFunctionBody(initStatements);
+      constructorBody = createFunctionBody([initStatement]);
     }
 
     return new FunctionExpression(tree.location, tree.name, false,
                                   constructorParams, null, [], constructorBody);
   }
 
-  /**
-   *  Appends assignment to this for the member variables after the super call.
-   */
-  appendInstanceInitializers_(constructor, initStatements, superClass) {
-    if (initStatements.length === 0) return constructor;
+  transformConstructorWithInitializer_(constructor, initExpression, superClass) {
+    if (superClass) {
+      let transformer = new SuperExpressionTransformer(initExpression);
+      return transformer.transformAny(constructor);
+    }
 
     let statements = constructor.body.statements;
-
-    if (superClass) {
-      let superExpressionIndex = -1;
-
-      for (let i = 0; i < statements.length; i++) {
-        let statement = statements[i];
-        if (statement.isDirectivePrologue()) {
-          continue;
-        }
-        if (isSuper(statement)) {
-          superExpressionIndex = i;
-          break;
-        }
-      }
-
-      statements = statements.slice();
-      statements.splice(superExpressionIndex + 1, 0, ...initStatements);
-    } else {
-      statements = prependStatements(statements, ...initStatements);
-    }
+    let initStatement = createExpressionStatement(initExpression);
+    statements = prependStatements(statements, initStatement);
 
     return new PropertyMethodAssignment(constructor.location, false,
         constructor.functionKind, constructor.name, constructor.parameterList,
@@ -485,32 +466,11 @@ export class ClassTransformer extends TempVarTransformer {
 
 // TODO(vicb): Does not handle computed properties
 function appendStaticInitializers(staticObject, initStaticMemberVars) {
-  // Initializes static member variables
   if (initStaticMemberVars.length === 0) return staticObject;
 
-  let properties =[];
-  for (let i = 0; i < initStaticMemberVars.length; i++) {
-    let mv = initStaticMemberVars[i];
-    properties.push(createPropertyNameAssignment(mv.name, mv.initializer));
-  }
+  let properties = initStaticMemberVars.map(
+      (mv) => createPropertyNameAssignment(mv.name, mv.initializer));
+
   return createObjectLiteralExpression(
       staticObject.propertyNameAndValues.concat(properties));
-}
-
-// TODO(vicb): Does not handle computed properties
-function getInstanceInitStatements(initInstanceVars) {
-    // Compute instance member variable initialization statements
-    let initStatements = [];
-    for (let i = 0; i < initInstanceVars.length; i++) {
-      let mv = initInstanceVars[i];
-      let name = mv.name.literalToken;
-      initStatements.push(parseStatement `this.${name} = ${mv.initializer};`);
-    }
-  return initStatements;
-}
-
-function isSuper(statement) {
-  return statement.type === EXPRESSION_STATEMENT &&
-         statement.expression.type === CALL_EXPRESSION &&
-         statement.expression.operand.type === SUPER_EXPRESSION;
 }
--- a/src/codegeneration/ES6ClassTransformer.js
+++ b/src/codegeneration/ES6ClassTransformer.js
@@ -1,0 +1,203 @@
+// Copyright 2015 Traceur Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the 'License');
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an 'AS IS' BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {
+  CONSTRUCTOR
+} from '../syntax/PredefinedName.js';
+import {
+  AnonBlock,
+  ClassExpression,
+  ClassDeclaration,
+  FormalParameterList,
+  PropertyMethodAssignment,
+} from '../syntax/trees/ParseTrees.js';
+import {
+  GET_ACCESSOR,
+  PROPERTY_METHOD_ASSIGNMENT,
+  PROPERTY_VARIABLE_DECLARATION,
+  SET_ACCESSOR,
+} from '../syntax/trees/ParseTreeType.js';
+import {TempVarTransformer} from './TempVarTransformer.js';
+import {
+  createBindingIdentifier,
+  createFunctionBody,
+  createIdentifierToken,
+  createImmediatelyInvokedFunctionExpression,
+  createLiteralPropertyName,
+  createRestParameter,
+} from './ParseTreeFactory.js';
+import {
+  parsePropertyDefinition,
+  parseStatement,
+} from './PlaceholderParser.js';
+import {propName} from '../staticsemantics/PropName.js';
+import {prependStatements} from './PrependStatements.js';
+import {
+  transformConstructor,
+  getInstanceInitExpression,
+} from './MemberVariableConstructorTransformer.js';
+
+/**
+ * Transform member variable declarations to valid ES6 statements.
+ *
+ * - instance variables are initialized in the constructor,
+ * - static variables are initialized after the class definition,
+ *   through `Object.defineProperty(...)`
+ */
+export class ES6ClassTransformer extends TempVarTransformer {
+  transformClassElements_(tree) {
+    let elements = [];
+    let initInstanceVars = [], initStaticVars = [];
+    let constructor;
+    let constructorIndex = 0;
+
+    tree.elements.forEach((tree) => {
+      let initVars;
+      if (tree.isStatic) {
+        initVars = initStaticVars;
+      } else {
+        initVars = initInstanceVars;
+      }
+
+      switch (tree.type) {
+        case GET_ACCESSOR:
+        case SET_ACCESSOR:
+          elements.push(this.transformAny(tree));
+          break;
+
+        case PROPERTY_METHOD_ASSIGNMENT:
+          if (!tree.isStatic && propName(tree) === CONSTRUCTOR) {
+            constructor = tree;
+            constructorIndex = elements.length;
+          } else {
+            elements.push(this.transformAny(tree));
+          }
+          break;
+
+        case PROPERTY_VARIABLE_DECLARATION:
+          tree = this.transformAny(tree);
+          if (tree.initializer !== null) {
+            initVars.push(tree);
+          }
+          break;
+
+        default:
+          throw new Error(`Unexpected class element: ${tree.type}`);
+      }
+    });
+
+    if (initInstanceVars.length > 0) {
+      let initExpression = getInstanceInitExpression(initInstanceVars);
+
+      if (!constructor) {
+        constructor = this.getDefaultConstructor_(tree);
+      }
+
+      constructor = transformConstructor(constructor, initExpression,
+          tree.superClass);
+    }
+
+    if (constructor) {
+      elements.splice(constructorIndex, 0, constructor);
+    }
+
+    return {
+      elements,
+      initStaticVars,
+    };
+  }
+
+  /**
+   * Transforms a single class declaration
+   *
+   * @param {ClassDeclaration} tree
+   * @return {ParseTree}
+   */
+  transformClassDeclaration(tree) {
+    let {
+      elements,
+      initStaticVars,
+    } = this.transformClassElements_(tree);
+
+    let superClass = this.transformAny(tree.superClass);
+    let classDecl = new ClassDeclaration(tree.location, tree.name, superClass,
+        elements, tree.annotations, tree.typeParameters);
+
+    if (initStaticVars.length === 0) {
+      return classDecl;
+    }
+
+    let statements = createStaticInitializerStatements(tree.name, initStaticVars);
+    statements = prependStatements(statements, classDecl);
+
+    return new AnonBlock(null, statements);
+  }
+
+  /**
+   * Transforms a single class expression
+   *
+   * @param {ClassExpression} tree
+   * @return {ParseTree}
+   */
+  transformClassExpression(tree) {
+    let {
+      elements,
+      initStaticVars,
+    } = this.transformClassElements_(tree);
+
+    let superClass = this.transformAny(tree.superClass);
+    let classExpression = new ClassExpression(tree.location, tree.name,
+        superClass, elements, tree.annotations, tree.typeParameters);
+
+    if (initStaticVars.length === 0) {
+      return classExpression;
+    }
+
+    this.pushTempScope();
+    let id = this.getTempIdentifier();
+    let className = createBindingIdentifier(id);
+    let statements = [
+      parseStatement `let ${id} = ${classExpression}`,
+      ...createStaticInitializerStatements(className, initStaticVars),
+      parseStatement `return ${className}`
+    ];
+    let body = createFunctionBody(statements);
+    this.popTempScope();
+
+    return createImmediatelyInvokedFunctionExpression(body);
+  }
+
+  getDefaultConstructor_(tree) {
+    if (tree.superClass) {
+      let param = createRestParameter(createIdentifierToken('args'));
+      let paramList = new FormalParameterList(null, [param]);
+      let body = createFunctionBody([parseStatement `super(...args)`]);
+      let name = createLiteralPropertyName(CONSTRUCTOR);
+      return new PropertyMethodAssignment(tree.location, false, null, name,
+          paramList, null, [], body);
+    }
+
+    return parsePropertyDefinition `constructor() {}`;
+  }
+}
+
+// TODO(vicb): Does not handle computed properties
+function createStaticInitializerStatements(className, initStaticMemberVars) {
+  return initStaticMemberVars.map((mv) => {
+    let propName = mv.name.literalToken.value;
+    return parseStatement
+        `Object.defineProperty(${className}, ${propName}, {enumerable: true,
+        configurable: true, value: ${mv.initializer}, writable: true})`;
+  });
+}
--- a/src/syntax/Parser.js
+++ b/src/syntax/Parser.js
@@ -324,6 +324,10 @@ const FUNCTION_STATE_DERIVED_CONSTRUCTOR = 1 << 5;
 const FUNCTION_STATE_GENERATOR = 1 << 6;
 const FUNCTION_STATE_ASYNC = 1 << 7;
 
+const FUNCTION_STATE_LENIENT =
+    FUNCTION_STATE_METHOD | FUNCTION_STATE_GENERATOR |
+    FUNCTION_STATE_ASYNC | FUNCTION_STATE_DERIVED_CONSTRUCTOR;
+
 /**
  * This is used to track the functions as the parser descends. It allows
  * us to determine if we are in a function and what kind of function it is.
@@ -938,9 +942,7 @@ export class Parser {
    */
   parseStatement() {
     // Allow return, yield and await.
-    let fs = this.pushFunctionState_(
-        FUNCTION_STATE_FUNCTION | FUNCTION_STATE_GENERATOR |
-        FUNCTION_STATE_ASYNC);
+    let fs = this.pushFunctionState_(FUNCTION_STATE_LENIENT);
     let result = this.parseModuleItem_(this.peekType_());
     this.popFunctionState_(fs);
     return result;
@@ -954,9 +956,7 @@ export class Parser {
    */
   parseStatements() {
     // Allow return, yield and await.
-    let fs = this.pushFunctionState_(
-        FUNCTION_STATE_FUNCTION | FUNCTION_STATE_GENERATOR |
-        FUNCTION_STATE_ASYNC);
+    let fs = this.pushFunctionState_(FUNCTION_STATE_LENIENT);
     let result = this.parseModuleItemList_();
     this.popFunctionState_(fs);
     return result;
@@ -2621,8 +2621,7 @@ export class Parser {
   }
 
   parseExpression() {
-    let fs = this.pushFunctionState_(FUNCTION_STATE_FUNCTION |
-        FUNCTION_STATE_GENERATOR | FUNCTION_STATE_ASYNC);
+    let fs = this.pushFunctionState_(FUNCTION_STATE_LENIENT);
     let expression = this.parseExpression_();
     this.popFunctionState_(fs);
     return expression;
--- a/test/unit/codegeneration/ES6ClassTransformer.js
+++ b/test/unit/codegeneration/ES6ClassTransformer.js
@@ -1,0 +1,178 @@
+// Copyright 2015 Traceur Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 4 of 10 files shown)


src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,232 @@
...
+        createClassDeclaration(tree.name, tree.superClass, elements));
+
+    return new AnonBlock(null, statements);
+  }
+
+  getDefaultConstructor_(tree, initStatements) {
+    let constructorParams = createEmptyParameterList();
+    let constructorBody;
+    if (tree.superClass) {
+      let statement = parseStatement `super(...arguments)`;
```
vicb: @arv this would trigger an error:
`null:1:1: super is only allowed in methods`

What would be the best way to handle this, create the super expression by hand ?

vicb:
I've just added some code for class expression.

@arv do you have some time this week for a quick chat about this.

The current (invalid) transform:

```
// source
// Options: --member-variables --output-language=es6 --script --source-maps 

class C {
  c1 = 1;
  c2;
  static c3 = 3;
}

var t = class {
  t1 = 1;
  t2;
  static t3 = 3;
}

// es6 output
class C {
  constructor() {
    this.c1 = 1;
  }
}
Object.defineProperty(C, "c3", {value: 3});
var t = (let $__0 = class {
  constructor() {
    this.t1 = 1;
  }
};, Object.defineProperty($__0, "t3", {value: 3});, "$__0";);
//# sourceMappingURL=traceured.js.map
```

src/codegeneration/ES6ClassTransformer.js
arv: Two options.

1. Change the Parser to be even more relaxed for this. I only made the relaxations that I needed at the time I added the FunctionState. What you need to do is add FUNCTION_STATE_DERIVED_CONSTRUCTOR.
2. Create the expression by hand.

vicb: something like
```js
  parseStatement() {
    // Allow return, yield and await.
    let fs = this.pushFunctionState_(
        FUNCTION_STATE_FUNCTION | FUNCTION_STATE_GENERATOR |
        FUNCTION_STATE_ASYNC | FUNCTION_STATE_DERIVED_CONSTRUCTOR);
    let result = this.parseModuleItem_(this.peekType_());
    this.popFunctionState_(fs);
    return result;
  }
```

vicb: It would not work because `super` as to be in a method: https://github.com/google/traceur-compiler/blob/980cc4bc04b820e38c571e4713bea9d9c559aa19/src/syntax/Parser.js#L1962

Should we fix `parserSuperExpression_()` or also add the `FUNCTION_STATE_METHOD` flag ?

arv: Replace FUNCTION_STATE_FUNCTION with F_S_METHOD.

I was thinking to maybe add a `pushLenientFunctionState_()` and use that for all the top level parse methods that are entry points, or add a constant for convenience.

```js
const FUNCTION_STATE_LENIENT =
    FUNCTION_STATE_METHOD | FUNCTION_STATE_GENERATOR |
    FUNCTION_STATE_ASYNC | FUNCTION_STATE_DERIVED_CONSTRUCTOR;
```

vicb:
The latest commit use an IIFE:
```js
class C {
  constructor() {
    this.c1 = 1;
  }
}
Object.defineProperty(C, "c3", {value: 3});
var t = (function() {
  let $__0 = class {
    constructor() {
      this.t1 = 1;
    }
  };
  Object.defineProperty($__0, "t3", {value: 3});
  return $__0;
})();
```

src/codegeneration/ES6ClassTransformer.js
vicb: Thanks for the quick reply, added `FUNCTION_STATE_LENIENT` as the 2nd commit.

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,270 @@
...
+    if (!constructor) {
+      constructor = this.getDefaultConstructor_(tree, initStatements);
+    } else {
+      if (this.options_.memberVariables) {
+        constructor = this.appendInstanceInitializers_(constructor,
+            initStatements, tree.superClass);
+      }
+    }
+
+    elements.push(constructor);
```
arv: Doesn't this put the constructor last? I would prefer if the order is kept and if we have to generated a constructor that it goes first.

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,270 @@
...
+    } = this.transformClassElements_(tree);
+
+    let classExpression = new ClassExpression(tree.location, tree.name,
+        tree.superClass, elements, tree.annotations, tree.typeParameters);
+
+    if (initStaticVars.length === 0) {
+      return classExpression;
+    }
+
+    let id = this.idGenerator_.generateUniqueIdentifier();
```
arv: This is where temp var transformer becomes handy. It allows you to return the temporary id when you are done with it.

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,270 @@
...
+    let className = createBindingIdentifier(id);
+    let statements = [];
```
arv:
```js
let statements = [
  parseStatement `let ${id} = ${classExpression}`),
  ...createStaticInitializerStatements(className, initStaticVars),
  parseStatement `return ${className}`
];
```

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,270 @@
...
+
+    if (superClass) {
+      let superExpressionIndex = -1;
+
+      for (let i = 0; i < statements.length; i++) {
+        let statement = statements[i];
+        if (statement.isDirectivePrologue()) {
+          continue;
+        }
+        if (isSuper(statement)) {
```
arv: I still think transforming `super()` would be cleaner. `super()` is an expression and there is really no requirement that it is the only expression in an expression statement.

src/syntax/Parser.js
```diff
@@ -954,9 +956,7 @@ export class Parser {
    */
   parseStatements() {
     // Allow return, yield and await.
-    let fs = this.pushFunctionState_(
-        FUNCTION_STATE_FUNCTION | FUNCTION_STATE_GENERATOR |
-        FUNCTION_STATE_ASYNC);
+    let fs = this.pushFunctionState_(FUNCTION_STATE_LENIENT);
```
arv: There is a parseExpression too that should use this...

src/codegeneration/ES6ClassTransformer.js
vicb: I'm reusing working code for now, I'll update after.

ref: `super(), this.x = <expr>, this.y = <expr>`

vicb:
I've integrated your feedback + add superclass transformation

src/codegeneration/ES6ClassTransformer.js
vicb: I've given this a try:

```js
      statements = statements.slice();
      let initExpression = createCommaExpression([
        statements[superExpressionIndex],
        ...initStatements
      ]);
      statements.splice(superExpressionIndex, 1, initExpression);
```

There is an issue with ";":
```
super();, this.c1 = 1;, this.c5 = 5;
```

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,272 @@
...
+            let pma = new PropertyMethodAssignment(tree.location, tree.isStatic,
+                tree.functionKind, tree.name, tree.parameterList,
+                tree.typeAnnotation, tree.annotations, tree.body,
+                tree.debugName);
+            elements.push(this.transformAny(pma));
+          }
+          break;
+
+        case PROPERTY_VARIABLE_DECLARATION:
+          this.transformAny(tree);
```
arv: Is this correct? Why don't you care about the result here?

The way I would do this is that I would make transformPropertyVariableDeclaration return an AssignmentExpression (this.x = expr).

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,272 @@
...
+          if (tree.initializer !== null) {
```
arv: I'm not sure if we should set it to undefined in this case?

src/codegeneration/ES6ClassTransformer.js
vicb: This should be addressed in the latest commit where I've added a proper transformer.

I was also able to do some refactoring / cleaning because we now know that ` super()`  is always here for derived class

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,272 @@
...
+          throw new Error(`Unexpected class element: ${tree.type}`);
+      }
+    });
+
+    let initStatements = getInstanceInitStatements(initInstanceVars);
+
+    if (!constructor) {
+      constructor = this.getDefaultConstructor_(tree, initStatements);
+    } else {
+      if (this.options_.memberVariables) {
```
arv: else if

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,272 @@
...
+    let {
+      elements,
+      initStaticVars,
+    } = this.transformClassElements_(tree);
+
+    let statements = createStaticInitializerStatements(tree.name, initStaticVars);
+    let superClass = this.transformAny(tree.superClass);
+
+    statements = prependStatements(statements,
+        createClassDeclaration(tree.name, superClass, elements));
```
arv: Does this drop annotations on the class declaration? Can you add a test?

src/codegeneration/ES6ClassTransformer.js
vicb: changed in the last commit

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,272 @@
...
+    let constructorParams = createEmptyParameterList();
+    let constructorBody;
+    if (tree.superClass) {
+      let statement = parseStatement `super(...arguments)`;
+      constructorBody = createFunctionBody([statement, ...initStatements]);
+    } else {
+      constructorBody = createFunctionBody(initStatements);
+    }
+
+    let name = new LiteralPropertyName(null, createIdentifierToken(CONSTRUCTOR));
```
arv: Can you use parsePropertyDefinition from PlaceholderParser.js

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,272 @@
...
+
+    return new PropertyMethodAssignment(tree.location, false,
+        null, name, constructorParams, null, [], constructorBody);
+  }
+
+  /**
+   *  Appends assignment to this for the member variables after the super call.
```
arv: -1 ws

src/codegeneration/ES6ClassTransformer.js
arv: Here is what I suggest.

Do another transformer, pass in the AssignmentExpressions you created earlier. Then call transformAny(constructor) (you might need to shortcut the transformer for class/object literals). This transformer would replace `super()` with `(super(), this.x = a, this.y = b)`.

This transformer can be used for the non es6 class transformer as well.

src/codegeneration/ES6ClassTransformer.js
vicb: sorry ? don't get it ?

src/codegeneration/ES6ClassTransformer.js
arv: The problem is that your assumption that `super()` is a statement is an issue. The following code is actually valid:

```js
class C extends B {
  constructor() {
    1 + super();
  }
}
```

Actually, this needs to be transformed into `(super(), this.x0 = e0, ... , this.xn = en, this)` as you can see by this V8 result:

```
$ out/x64.release/d8 -e "'use strict'; class B {} class D extends B { constructor() { print(super()); } }; new D"
[object Object]
```

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,272 @@
...
+    let statement = parseStatement
+        `Object.defineProperty(${className}, ${propName}, {value: ${mv.initializer}});`;
+    statements.push(statement);
+  }
+  return statements;
+}
+
+// TODO(vicb): Does not handle computed properties
+function getInstanceInitStatements(initInstanceVars) {
+    // Compute instance member variable initialization statements
```
arv: wrong indentation

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,272 @@
...
+}
+
+// TODO(vicb): Does not handle computed properties
+function createStaticInitializerStatements(className, initStaticMemberVars) {
+  let statements = [];
+  for (let i = 0; i < initStaticMemberVars.length; i++) {
+    let mv = initStaticMemberVars[i];
+    let propName = mv.name.literalToken.value;
+    let statement = parseStatement
+        `Object.defineProperty(${className}, ${propName}, {value: ${mv.initializer}});`;
```
arv: This needs `configurable: true, enumerable: true, writable: true` too :'(

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    return new PropertyMethodAssignment(constructor.location, false,
+        constructor.functionKind, constructor.name, constructor.parameterList,
+        constructor.typeAnnotation, constructor.annotations,
+        createFunctionBody(statements));
+  }
+}
+
+// TODO(vicb): Does not handle computed properties
```
arv: Now that it is using defineProperty it is trivial to support computed properties.

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,272 @@
...
+    let mv = initStaticMemberVars[i];
+    let propName = mv.name.literalToken.value;
+    let statement = parseStatement
+        `Object.defineProperty(${className}, ${propName}, {value: ${mv.initializer}});`;
+    statements.push(statement);
+  }
+  return statements;
+}
+
+// TODO(vicb): Does not handle computed properties
```
arv: This is easy too... change `this.x` into `this['x']`. Maybe extract the code to convert the property name into an expression from the existing code.... this can all be done in a follow up PR of course.

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,261 @@
...
+      let initVars;
+      if (tree.isStatic) {
+        initVars = initStaticVars;
+      } else {
+        initVars = initInstanceVars;
+      }
+
+      switch (tree.type) {
+        case GET_ACCESSOR:
+          let getter = new GetAccessor(tree.location, tree.isSatic, tree.name,
```
arv: typo isSatic. Is there a missing test?

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,261 @@
...
+              tree.typeAnnotation, tree.annotations, tree.body);
+          elements.push(this.transformAny(getter));
+          break;
+
+        case SET_ACCESSOR:
+          let setter = new SetAccessor(tree.location, tree.isSatic, tree.name,
```
arv: same here

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,261 @@
...
+              tree.parameterList, tree.annotations, tree.body);
+          elements.push(this.transformAny(setter));
+          break;
+
+        case PROPERTY_METHOD_ASSIGNMENT:
+          if (!tree.isStatic && propName(tree) === CONSTRUCTOR) {
```
arv: Maybe turn this into a helper function? I feel like we do this check in a few places already.

You also need to check functionKind.

arv: Maybe not actually, in V8 I check this in the parser but I think I forgot to add these checks to Traceur so Traceur is wrong to parse such a class.

I filed #1793 so you do not have to care about this at this stage.

src/codegeneration/ES6ClassTransformer.js
arv:
```js
/**
 * Appends  ..
 */
```

ws = whitespace... there was one too many before Appends

src/syntax/Parser.js
```diff
@@ -2247,7 +2247,8 @@ export class Parser {
     let name = this.parsePropertyName_();
 
     if (this.options_.propertyMethods && this.peek_(OPEN_PAREN)) {
-      let fs = this.pushFunctionState_(FUNCTION_STATE_METHOD);
+      let fs = this.pushFunctionState_(FUNCTION_STATE_METHOD |
+          FUNCTION_STATE_DERIVED_CONSTRUCTOR);
```
arv: This doesn't look right. I assume this is a left over from when you were trying to make the placeholder parser to work.

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,265 @@
...
+    ];
+    let body = createFunctionBody(statements);
+    this.popTempScope();
+
+    return createImmediatelyInvokedFunctionExpression(body);
+  }
+
+  getDefaultConstructor_(tree) {
+    if (tree.superClass) {
+      return parsePropertyDefinition `constructor() {
```
arv: use:

```js
constructor(...args) {
  super(...args);
}
```

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,265 @@
...
+      ...createStaticInitializerStatements(className, initStaticVars),
+      parseStatement `return ${className}`
+    ];
+    let body = createFunctionBody(statements);
+    this.popTempScope();
+
+    return createImmediatelyInvokedFunctionExpression(body);
+  }
+
+  getDefaultConstructor_(tree) {
```
arv: maybe change this to hasSuper since you really do not care about tree in here

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,265 @@
...
+
+class SuperExpressionTransformer extends ParseTreeTransformer {
+  constructor(expression) {
+    super();
+    this.expression = expression;
+  }
+
+  transformCallExpression(tree) {
+    if (tree.operand.type === SUPER_EXPRESSION) {
+      var thisExpression = parseExpression `this`;
```
arv: createThisExpression?

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,265 @@
...
+      return createParenExpression(
+          createCommaExpression([tree, this.expression, thisExpression]));
+    }
+
+    return tree;
+  }
```
arv: I'm little bit concerned about nested classes:

```js
class A extends B {
  x = 42;
  constructor() {
    super();
    class C extends D {
      constructor() {
        super(); // Ooops.
      }
  }
}
```


I think the simplest solution is just to add:

transformClassDeclaration(tree) { return tree; }
transformClassExpression(tree) { return tree; }

src/syntax/Parser.js
vicb: Nope, it wouldn't work without this ("super only allowed in derived ctor")

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,265 @@
...
+    tree.elements.forEach((tree) => {
+      let initVars;
+      if (tree.isStatic) {
+        initVars = initStaticVars;
+      } else {
+        initVars = initInstanceVars;
+      }
+
+      switch (tree.type) {
+        case GET_ACCESSOR:
```
arv: This can be written as:

```js
case GET_ACCESSOR:
case SET_ACCESSOR:
  elements.push(this.transformAny(tree));
  break;
```

There is no need to create that extra tree here.

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,265 @@
...
+              tree.parameterList, tree.annotations, tree.body);
+          elements.push(this.transformAny(setter));
+          break;
+
+        case PROPERTY_METHOD_ASSIGNMENT:
+          if (!tree.isStatic && propName(tree) === CONSTRUCTOR) {
+            constructor = tree;
+            constructorIndex = elements.length;
+          } else {
+            let pma = new PropertyMethodAssignment(tree.location, tree.isStatic,
```
arv: Same here... 

```js
elements.push(this.transformAny(tree));
```

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,265 @@
...
+                tree.functionKind, tree.name, tree.parameterList,
+                tree.typeAnnotation, tree.annotations, tree.body,
+                tree.debugName);
+            elements.push(this.transformAny(pma));
+          }
+          break;
+
+        case PROPERTY_VARIABLE_DECLARATION:
+          tree = this.transformAny(tree);
```
arv: add a transformPropertyVariableDeclaration that handles this transformatio instead of using getInstanceInitExpression.

```js
transformPropertyVariableDeclaration(tree) {
  return new BinaryExpression(tree.location, 
      new MemberExpression(...),
      EQUAL,
      this.transformAny(tree.initializer))
}
```

// pseudo code

arv: Maybe not... since you need to do different things for instance vs static.

vicb:
@arv the last commit add test however I don't understand why the code is parsed as an expression, the output is:

``` 
      + expected - actual

      +(function() {
      +  let $__0 = class C {
      +    constructor() {
      +      this.c1 = 1, this.c2 = 2;
      +    }
      +  };
      +  Object.defineProperty($__0, "c0", {
      +    value: 0,
      +    enumerable: true,
      +    configurable: true,
      +    writable: true
      +  });
      +  return $__0;
      +})()
      -...
```

src/syntax/Parser.js
arv: This change would break the following... I guess I missed a test for it?

```js
class C {
  m() {
    super();  // Should be SyntaxError
  }
}
```

I guess that means that we cannot use parsePropertyDefinition for this :(

arv:
Make sure that you are creating valid trees. You can test this with `options.validate = true`.

src/syntax/Parser.js
vicb: I'll revert the changes

vicb:
The test was using `parseExpression()`, fixed

vicb:
Travis as one error with annotations: https://travis-ci.org/google/traceur-compiler/builds/53180488#L210
I don't know how to fix this.

I have two more errors locally:
```
  2) SourceMapMapping columnIndexGreaterOrEqual:
     Error: timeout of 2000ms exceeded
      at [object Object].<anonymous> (/home/victor/dart/traceur-compiler/node_modules/mocha/lib/runnable.js:139:19)
      at Timer.listOnTimeout [as ontimeout] (timers.js:110:15)

  3) SourceMapMapping testNextPosition:
     Error: timeout of 2000ms exceeded
      at [object Object].<anonymous> (/home/victor/dart/traceur-compiler/node_modules/mocha/lib/runnable.js:139:19)
      at Timer.listOnTimeout [as ontimeout] (timers.js:110:15)
```

src/codegeneration/ES6ClassTransformer.js
vicb: fixed

src/codegeneration/ES6ClassTransformer.js
vicb: see https://github.com/google/traceur-compiler/issues/1795

src/codegeneration/ES6ClassTransformer.js
vicb: fixed

src/codegeneration/ClassTransformer.js
```diff
@@ -252,10 +251,10 @@ export class ClassTransformer extends TempVarTransformer {
     let initStatements = getInstanceInitStatements(initInstanceVars);
     let func;
 
-    if (!hasConstructor) {
+    if (!constructor) {
       func = this.getDefaultConstructor_(tree, internalName, initStatements);
     } else {
-      if (this.options_.memberVariables) {
+      if (initInstanceVars.length > 0) {
```
arv: It is not clear to me why you no longer check the option here?

src/codegeneration/ClassTransformer.js
```diff
@@ -448,34 +447,19 @@ export class ClassTransformer extends TempVarTransformer {
...
-          break;
-        }
-      }
-
-      statements = statements.slice();
-      statements.splice(superExpressionIndex + 1, 0, ...initStatements);
-    } else {
-      statements = prependStatements(statements, ...initStatements);
+      let initExpression = createCommaExpression(
+          initStatements.map((statement) => statement.expression));
```
arv: Why the map? Can you just store expressions instead of expression statements?

src/codegeneration/ClassTransformer.js
```diff
@@ -448,34 +447,19 @@ export class ClassTransformer extends TempVarTransformer {
                                   constructorParams, null, [], constructorBody);
   }
 
-  /**
-   *  Appends assignment to this for the member variables after the super call.
-   */
   appendInstanceInitializers_(constructor, initStatements, superClass) {
```
arv: Maybe rename this to indicate that it is transforming the constructor?

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,259 @@
...
+} from './PlaceholderParser.js';
+import {propName} from '../staticsemantics/PropName.js';
+import {prependStatements} from './PrependStatements.js';
+
+/**
+ * Transform member variable declarations to valid ES6 statements.
+ *
+ * - instance variables are initialized in the constructor,
+ * - static variables are initialized after the class definition,
+ *   through ` Object.defineProperty(...)`
```
arv: remove a space after `

src/codegeneration/ES6ClassTransformer.js
arv: I noticed that the ClassTransformer uses 

```js
this.transformAny(tree);
```

and throwing away the result (like this did before). Can you fix that too?

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,259 @@
...
+        default:
+          throw new Error(`Unexpected class element: ${tree.type}`);
+      }
+    });
+
+    if (initInstanceVars.length > 0) {
+      let initExpression = getInstanceInitExpression(initInstanceVars);
+
+      if (!constructor) {
+        constructor = this.getDefaultConstructor_(!!tree.superClass);
```
arv: no need for !! since this is all private

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,259 @@
...
+   * @param {ClassDeclaration} tree
+   * @return {ParseTree}
+   */
+  transformClassDeclaration(tree) {
+    let {
+      elements,
+      initStaticVars,
+    } = this.transformClassElements_(tree);
+
+    let statements = createStaticInitializerStatements(tree.name, initStaticVars);
```
arv: For the ClassExpression you checked the length of initStaticVars. Maybe do the same here.

src/codegeneration/ClassTransformer.js
vicb: that's probably something we could do - currently we use initStatements below (l 461) and for getDefaultCtor

src/codegeneration/ES6ClassTransformer.js
vicb: will fix

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,259 @@
...
+    this.popTempScope();
+
+    return createImmediatelyInvokedFunctionExpression(body);
+  }
+
+  getDefaultConstructor_(hasSuper) {
+    if (hasSuper) {
+      var pma = parsePropertyDefinition `constructor(...args) {}`;
+      // `parsePropertyDefinition()` would not allow `super(...)` in method body
+      pma.body.statements = [parseStatement `super(...args)`];
```
arv: ParseTrees are immutable by a gentlemen's agreement. By mutating it the caching will break.

I think for now you need to create a new tree (or fix parsePropertyDefinition).

src/codegeneration/ClassTransformer.js
vicb: will fix

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,259 @@
...
+
+// TODO(vicb): Does not handle computed properties
+function createStaticInitializerStatements(className, initStaticMemberVars) {
+  let statements = [];
+  for (let i = 0; i < initStaticMemberVars.length; i++) {
+    let mv = initStaticMemberVars[i];
+    let propName = mv.name.literalToken.value;
+    let statement = parseStatement
+        `Object.defineProperty(${className}, ${propName}, {value: ${mv.initializer},
+        enumerable: true, configurable: true, writable: true})`;
```
arv: keep these in alphabetical order. Then V8 can share the internal hidden class.

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,259 @@
...
+    statements.push(statement);
+  }
+  return statements;
+}
+
+// TODO(vicb): Does not handle computed properties
+function getInstanceInitExpression(initInstanceVars) {
+    let expressions = [];
```
arv: wrong indentation

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,259 @@
...
+    for (let i = 0; i < initInstanceVars.length; i++) {
```
arv: or use map()

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,259 @@
...
+      let mv = initInstanceVars[i];
+      let name = mv.name.literalToken;
```
arv: Good, we are keeping the location this way.

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  transformCallExpression(tree) {
+    if (tree.operand.type === SUPER_EXPRESSION) {
+      var thisExpression = createThisExpression();
+      return createParenExpression(
+          createCommaExpression([tree, this.expression, thisExpression]));
+    }
+
+    return tree;
```
arv: return super.transformCallExpression(tree);

or `f(super())` will fail.

test/unit/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,185 @@
...
+      memberVariables: false,
+      annotations: true,
+      validate: true,
+    });
+    var parser = new Parser(file, null, options);
+    var transformed = parser.parseScript();
+    return write(transformed);
+  }
+
+  function parseScript(content) {
```
arv: pass in options so we use the same one in both cases?

test/unit/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,185 @@
...
+
+  function normalizeScript(content) {
+    var file = new SourceFile('test', content);
+    var options = new Options({
+      memberVariables: false,
+      annotations: true,
+      validate: true,
+    });
+    var parser = new Parser(file, null, options);
+    var transformed = parser.parseScript();
```
arv: this tree is not transformed, rename?

test/unit/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,185 @@
...
+
+  function parseScript(content) {
+    var file = new SourceFile('test', content);
+    var options = new Options({
+      memberVariables: true,
+      annotations: true,
+      validate: true,
+    });
+    var parser = new Parser(file, null, options);
+    var tree = parser.parseScript();
```
arv: refactor to reduce code duplication?

arv:
I think this looks good. However, isn't there a lot of code being duplicated between ClassTransformer and ES6ClassTransformer? Also, maybe we could always run ES6ClassTransformer (and rename it) before ClassTransformer? Then ClassTransformer does not need to know about member variables. Maybe that is problematic because we lose context needed for annotations?

test/unit/codegeneration/ES6ClassTransformer.js
vicb: done

vicb:
> However, isn't there a lot of code being duplicated between ClassTransformer and ES6ClassTransformer?

I thought about that but:
- the code is slighty different,
- the `ClassTransformer` is already a huge beast, adding ESx switches would make the logic hard to follow.

> Also, maybe we could always run ES6ClassTransformer (and rename it) before ClassTransformer?

May be. I have to think about it.

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,262 @@
...
+// TODO(vicb): Does not handle computed properties
+function getInstanceInitExpression(initInstanceVars) {
+  let expressions = initInstanceVars.map((mv) => {
+    let name = mv.name.literalToken;
+    return parseExpression `this.${name} = ${mv.initializer}`;
+  });
+  return createCommaExpression(expressions);
+}
+
+export class SuperExpressionTransformer extends ParseTreeTransformer {
```
arv: Can you move this to its own file and rename it so that it is clear that this is used for member variables?

arv: MemberVariableSuperCallTransformer?

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,262 @@
...
+  return initStaticMemberVars.map((mv) => {
+    let propName = mv.name.literalToken.value;
+    return parseStatement
+        `Object.defineProperty(${className}, ${propName}, {enumerable: true,
+        configurable: true, value: ${mv.initializer}, writable: true})`;
+  });
+}
+
+// TODO(vicb): Does not handle computed properties
+function getInstanceInitExpression(initInstanceVars) {
```
arv: This must be same between the two class transformers?

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,262 @@
...
+      let body = createFunctionBody([parseStatement `super(...args)`]);
+      let name = createLiteralPropertyName(CONSTRUCTOR);
+      return new PropertyMethodAssignment(tree.location, false, null, name,
+          paramList, null, [], body);
+    }
+
+    return parsePropertyDefinition `constructor() {}`;
+  }
+
+  transformConstructorWithInitializer_(constructor, initExpression, superClass) {
```
arv: This can be shared?

src/codegeneration/ES6ClassTransformer.js
```diff
@@ -0,0 +1,262 @@
...
+  createParenExpression,
+  createThisExpression,
+} from './ParseTreeFactory.js';
+import {
+  parseExpression,
+  parsePropertyDefinition,
+  parseStatement,
+} from './PlaceholderParser.js';
+import {propName} from '../staticsemantics/PropName.js';
+import {prependStatements} from './PrependStatements.js';
```
arv: once done you can try this on the command line

```
$ node tools/find-unused-imports.js src/codegeneration/ES6ClassTransformer.js 
```

or more fancy:

```
git diff master --name-only | xargs node tools/find-unused-imports.js
```

src/codegeneration/ES6ClassTransformer.js
vicb: it is, will factorize

src/codegeneration/ES6ClassTransformer.js
vicb: Super useful, thanks.
I've pushed the changes. Let me know what you think.

arv:
The Travis error looks real. You probably forgot to git add the file.

vicb:
Oops !

arv:
LGTM. Can you squash?

PR #1812, opened 2015-03-12
Fix template literals precedence for CallNew

### LINKED ISSUE #1811: templates: more or less the same as chromium:450942
The following code (from https://crrev.com/996223003/):

```js
(function testCallNew2() {
  "use strict";
  var log = [];
  function tag(x) {
    log.push(x);
    if (!(this instanceof tag)) {
      return tag;
    }
    this.x = x === void 0 ? null : x;
    return this;
  }
  var instance = new tag`x``y``z`;
  console.log("instance instanceof tag: " + instance instanceof tag);
  console.log("tag.prototype === instance.__proto__: " +
              (tag.prototype === Object.getPrototypeOf(instance)) );
  console.log(JSON.stringify(log));
})();
```

results in `new tag($__0)($__1)($__2);`, as opposed to `new(tag($__0)($__1)($__2));` (which would match the behaviour of Firefox and the above referenced patch). Similarly, also fails the original repro case for [the bug](https://crbug.com/450942).

Someone who is better at reading the spec BNF than me should be able to clarify the appropriate behaviour, but for now I'm leaning towards matching what Firefox is doing.

caitp:
Closes #1811

### DIFF
--- a/src/syntax/Parser.js
+++ b/src/syntax/Parser.js
@@ -3258,6 +3258,8 @@ export class Parser {
         let args = null;
         if (this.peek_(OPEN_PAREN)) {
           args = this.parseArguments_();
+        } else if (this.peek_(NO_SUBSTITUTION_TEMPLATE) || this.peek_(TEMPLATE_HEAD)) {
+          operand = this.parseMemberExpression_();
         }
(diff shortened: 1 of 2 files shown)


caitp:
mm, don't think this works :(

PR #1813, opened 2015-03-12
Parenthesize complex operands in NewExpression

### LINKED ISSUE #1811: templates: more or less the same as chromium:450942

caitp:
Closes #1811

### DIFF
--- a/src/codegeneration/TemplateLiteralTransformer.js
+++ b/src/codegeneration/TemplateLiteralTransformer.js
@@ -16,11 +16,12 @@ import {
   BINARY_EXPRESSION,
   COMMA_EXPRESSION,
   CONDITIONAL_EXPRESSION,
-  TEMPLATE_LITERAL_PORTION
+  TEMPLATE_LITERAL_PORTION,
+  TEMPLATE_LITERAL_EXPRESSION
 } from '../syntax/trees/ParseTreeType.js';
 import {
   LiteralExpression,
-  ParenExpression
+  NewExpression
 } from '../syntax/trees/ParseTrees.js';
 import {LiteralToken} from '../syntax/LiteralToken.js';
 import {ParseTreeTransformer} from './ParseTreeTransformer.js';
@@ -39,7 +40,8 @@ import {
   createCallExpression,
   createIdentifierExpression,
   createOperatorToken,
-  createStringLiteral
+  createStringLiteral,
+  createParenExpression
 } from './ParseTreeFactory.js';
 import {parseExpression} from './PlaceholderParser.js';
 
@@ -198,6 +200,21 @@ export class TemplateLiteralTransformer extends TempVarTransformer {
     return ParseTreeTransformer.prototype.transformFunctionBody.call(this, tree);
   }
 
+  transformNewExpression(tree) {
+    if (tree.operand) {
+      let operand = tree.operand;
+      do {
+        // If operand contains a tagged TemplateLiteral, parenthesize the entire
+        // operand.
+        if (operand.type === TEMPLATE_LITERAL_EXPRESSION && operand.operand !== null) {
+          tree = new NewExpression(tree.location, createParenExpression(tree.operand), tree.args);
+          break;
+        }
+      } while (operand = operand.operand);
+    }
+    return super.transformNewExpression(tree);
+  }
+
   transformTemplateLiteralExpression(tree) {
     if (!tree.operand)
       return this.createDefaultTemplateLiteral(tree);
@@ -232,7 +249,7 @@ export class TemplateLiteralTransformer extends TempVarTransformer {
         // Fall through.
       case COMMA_EXPRESSION:
       case CONDITIONAL_EXPRESSION:
-        return new ParenExpression(null, transformedTree);
+        return createParenExpression(transformedTree);
     }
 
     return transformedTree;
@@ -269,6 +286,6 @@ export class TemplateLiteralTransformer extends TempVarTransformer {
                                               transformedTree);
     }
 
-    return new ParenExpression(null, binaryExpression);
+    return new createParenExpression(binaryExpression);
   }
 }
--- a/test/feature/TemplateLiterals/NewExpression.js
+++ b/test/feature/TemplateLiterals/NewExpression.js
@@ -1,0 +1,44 @@
+// Options: --block-binding
+
+{
+  let i = 0, log = [];
+  this.logger = function(...e) {
+    log.push(e.join(' '));
+  }
+
+  // https://crbug.com/450942
+  assert.equal(typeof new Function`logger('a')`, 'object');
+  assert.deepEqual(log, ['a']);
+
+  log.length = 0;
+  function tag(...e) {
+    var text = e[0] && String.raw(...e);
+    if (this instanceof tag) {
+      log.push('new;' + text);
+    } else {
+      log.push('tag;' + text);
+      return tag;
+    }
+  }
+
+  assert.equal(typeof new tag`a``b``c`, 'object');
+  assert.deepEqual(log, [
+    'tag;a',
+    'tag;b',
+    'tag;c',
+    'new;undefined'
+  ]);
+
+  log.length = 0;
+  function C(cs) {
+    log.push(cs[0]);
+    if (this instanceof C) {
+      this.name = cs;
+    } else {
+      return C;
+    }
+  }
+
+  assert.deepEqual(new C`a``b``c`('test'), { name: 'test' });
+  assert.deepEqual(log, ['a', 'b', 'c', 't']);
+}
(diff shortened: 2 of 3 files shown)


caitp:
This one seems to work better :<

src/outputgeneration/ParseTreeWriter.js
```diff
@@ -1006,9 +1006,12 @@ export class ParseTreeWriter extends ParseTreeVisitor {
    * @param {NewExpression} tree
    */
   visitNewExpression(tree) {
+    var complexOperand = tree.operand.operand !== void 0;
```
arv: The writer should not have logic on it.

We never use undefined for valid properties. The right approach is to check the type.

So, if I parse

```js
new a.b.c()
```

Then the operand is `a.b.c` and you now want to add parens around that? That does not seem correct.

I think we need to look at the AST we generate. Is it correct? I think it is. However, our transformer is not aware of `new`.

```js
new a `b` `c`
```

is transformed into

```js
new a($__0)($__1)
```

when we need to transform it into

```js
new (a($__0)($__1))
```

test/feature/TemplateLiterals/NewExpression.js
```diff
@@ -0,0 +1,44 @@
+// Options: --block-binding
+
+{
+  let i = 0, log = [];
+  this.logger = function(...e) {
+    log.push(e.join(" "));
```
arv: Traceur uses Google's JS style guide which means that we prefer single quotes.

src/outputgeneration/ParseTreeWriter.js
caitp: Right now, it's relying on the vm's own idea of precedence, but I think the parser is mostly getting it right, so always parenthesizing the operand should be fine (just produces some ugly code).

Maybe it makes more sense to cover "complex" operands in transformNewExpression()

arv: The writer is supposed to just write the tree. It is supposed to round trip.

caitp: I get that, I'm just trying to figure out how to do this in ParseTreeTransformer =)

arv: We need to add a transformNewExpression to TemplateLiteralTransformer. Not yet sure how to best detect that the operand contains a template literal.

src/codegeneration/TemplateLiteralTransformer.js
```diff
@@ -198,6 +199,21 @@ export class TemplateLiteralTransformer extends TempVarTransformer {
...
+    if (tree.operand) {
+      var operand = tree.operand;
+      do {
+        // If operand contains a tagged TemplateLiteral, parenthesize the entire
+        // operand.
+        if (operand.operand && operand.type === TEMPLATE_LITERAL_EXPRESSION) {
+          tree.operand = new ParenExpression(null, tree.operand);
+          break;
+        }
+      } while (operand = operand.operand);
```
caitp: New approach, hows that?

src/codegeneration/TemplateLiteralTransformer.js
```diff
@@ -198,6 +199,21 @@ export class TemplateLiteralTransformer extends TempVarTransformer {
...
+    }
+    return ParseTreeTransformer.prototype.transformNewExpression.call(this, tree);
```
arv: Use `super.transformNewExpression` (I see that there is some old code above that uses ES5 style super)

src/codegeneration/TemplateLiteralTransformer.js
```diff
@@ -16,7 +16,8 @@ import {
   BINARY_EXPRESSION,
   COMMA_EXPRESSION,
   CONDITIONAL_EXPRESSION,
-  TEMPLATE_LITERAL_PORTION
+  TEMPLATE_LITERAL_PORTION,
```
arv: not used

src/codegeneration/TemplateLiteralTransformer.js
```diff
@@ -198,6 +199,21 @@ export class TemplateLiteralTransformer extends TempVarTransformer {
     return ParseTreeTransformer.prototype.transformFunctionBody.call(this, tree);
   }
 
+  transformNewExpression(tree) {
+    if (tree.operand) {
+      var operand = tree.operand;
```
arv: use `let`

src/codegeneration/TemplateLiteralTransformer.js
```diff
@@ -198,6 +199,21 @@ export class TemplateLiteralTransformer extends TempVarTransformer {
...
+      do {
+        // If operand contains a tagged TemplateLiteral, parenthesize the entire
+        // operand.
+        if (operand.operand && operand.type === TEMPLATE_LITERAL_EXPRESSION) {
```
arv: needs a type check, accessing non existing properties is something we try to not do.

src/codegeneration/TemplateLiteralTransformer.js
```diff
@@ -198,6 +199,21 @@ export class TemplateLiteralTransformer extends TempVarTransformer {
...
+          tree.operand = new ParenExpression(null, tree.operand);
```
arv: trees are immutable (but we do not actually freeze the for performance reasons)

src/codegeneration/TemplateLiteralTransformer.js
arv: This approach is solid.

src/codegeneration/TemplateLiteralTransformer.js
caitp: I'm not sure how to best do that, since `operand = operand.operand` is going to happen when continuing the loop

arv: Actually, just change it to:

```js
if (operand.type === TEMPLATE_LITERAL_EXPRESSION && operand.operand !== null) {
...
```

src/codegeneration/TemplateLiteralTransformer.js
caitp: it looks like it's used in this file to me

arv: Sorry, my bad... I got tricked by the diff.

src/codegeneration/TemplateLiteralTransformer.js
```diff
@@ -198,6 +200,21 @@ export class TemplateLiteralTransformer extends TempVarTransformer {
...
   }
 
+  transformNewExpression(tree) {
+    if (tree.operand) {
+      let operand = tree.operand;
+      do {
+        // If operand contains a tagged TemplateLiteral, parenthesize the entire
+        // operand.
+        if (operand.type === TEMPLATE_LITERAL_EXPRESSION && operand.operand !== null) {
+          tree = createNewExpression(createParenExpression(tree.operand), tree.args);
```
arv: This works but in general it is better to use the constructors so that the location is not lost.

```js
tree = new NewExpression(tree.location,
    createParenExpression(tree.operand), tree.args);
```

since the paren expression is generated code there is no location for it so either `createParenExpression(expr)` or `new ParenExpression(null, expr)` is fine.

src/codegeneration/TemplateLiteralTransformer.js
```diff
@@ -198,6 +200,21 @@ export class TemplateLiteralTransformer extends TempVarTransformer {
...
   }
 
+  transformNewExpression(tree) {
+    if (tree.operand) {
+      let operand = tree.operand;
+      do {
+        // If operand contains a tagged TemplateLiteral, parenthesize the entire
+        // operand.
+        if (operand.type === TEMPLATE_LITERAL_EXPRESSION && operand.operand !== null) {
+          tree = new NewExpression(tree.location, createParenExpression(tree.operand), tree.args);
```
caitp: done

arv:
LGTM

Can you add yourself to AUTHORS too?

caitp:
yeah, done

arv:
great... landing

PR #1824, opened 2015-03-16
Support supplemental code points while scanning

caitp:
since the unicode tables deal with supplemental codepoints, it is probably a good idea to actually scan them!

good things to do: for identifiers, use a generated identifier in codegen instead, so that it might work in legacy engines.

### DIFF
--- a/src/syntax/Scanner.js
+++ b/src/syntax/Scanner.js
@@ -195,7 +195,8 @@ function isRegularExpressionFirstChar(code) {
 }
 
 let index, input, length, token, lastToken, lookaheadToken, currentCharCode,
-    lineNumberTable, errorReporter, currentParser, options;
+    nextCharCode, lineNumberTable, errorReporter, currentParser, options,
+    currentCodePoint;
 
 /**
  * Scans javascript source code into tokens. All entrypoints assume the
@@ -291,7 +292,7 @@ export class Scanner {
     lastToken = null;
     token = null;
     lookaheadToken = null;
-    updateCurrentCharCode();
+    updateCurrentCodePoint();
   }
 
   get index() {
@@ -323,7 +324,7 @@ function nextRegularExpressionLiteralToken() {
 
   // body
   // Also, check if we are done with the body in case we had /=/
-  if (!(token.type === SLASH_EQUAL && currentCharCode === 47) &&
+  if (!(token.type === SLASH_EQUAL && currentCodePoint === 47) &&
       !skipRegularExpressionBody()) {
     return new LiteralToken(REGULAR_EXPRESSION,
                             getTokenString(beginIndex),
@@ -331,7 +332,7 @@ function nextRegularExpressionLiteralToken() {
   }
 
   // separating /
-  if (currentCharCode !== 47) {  // /
+  if (currentCodePoint !== 47) {  // /
     reportError('Expected \'/\' in regular expression literal');
     return new LiteralToken(REGULAR_EXPRESSION,
                             getTokenString(beginIndex),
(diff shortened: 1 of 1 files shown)


src/syntax/Scanner.js
```diff
@@ -1172,11 +1173,30 @@ function isAtEnd() {
...
   index++;
-  updateCurrentCharCode();
+  if (currentCodePoint >= 0x10000) {
+    index++;
+  }
+  updateCurrentCodePoint();
+}
+
+function updateCurrentCodePoint() {
+  currentCodePoint = currentCharCode = input.charCodeAt(index);
```
arv: is currentCharCode still used?

src/syntax/Scanner.js
```diff
@@ -1172,11 +1173,30 @@ function isAtEnd() {
...
+  nextCharCode = input.charCodeAt(index + 1) | 0;
```
arv: This is really hot code. What is the perf impact?

arv: nextCharCode is only used in this function. Use a local variable.

src/syntax/Scanner.js
```diff
@@ -1172,11 +1173,30 @@ function isAtEnd() {
...
+  if (isLeadSurrogate(currentCodePoint) && isTrailSurrogate(nextCharCode)) {
```
arv: One simple optimization would probably be to to check if the currentCodePoint is ASCII.

```js
function updateCurrentCodePoint() {
  currentCodePoint = input.charCodeAt(index);
  if (currentCodePoint < 128) return;
  ...
}
```

mathiasbynens:
Small stylistic nit: how about using consistently using hexadecimal integer literals to refer to code points?

E.g.:

```js
// instead of…
currentCodePoint === 47
// use:
currentCodePoint === 0x2F
```

This makes it much clearer that this is about U+002F, IMHO.

arv:
@mathiasbynens Maybe unicode geeks like you dream in hex but I still operate better with decimals.

mathiasbynens:
@arv It’s just inconsistent right now. See `0xD800` at <https://github.com/google/traceur-compiler/pull/1824/files#diff-05e5b7795cfe69ccb6421b981250effbR1191> and `0xDC00` at <https://github.com/google/traceur-compiler/pull/1824/files#diff-05e5b7795cfe69ccb6421b981250effbR1195>.

caitp:
I would recognize 0xFFFE easier than 65534, but that may not apply to decimals in the ascii range

mathiasbynens:
@caitp Pfff, double standards! :P

src/syntax/Scanner.js
```diff
@@ -1172,11 +1172,33 @@ function isAtEnd() {
...
+    // If code point is higher than Plane 0, move past trail surrogate
+    index++;
+  }
+  updateCurrentCodePoint();
+}
+
+function updateCurrentCodePoint() {
+  currentCodePoint = input.charCodeAt(index);
+  // Return if code point is in Plane 0
+  if (currentCodePoint < 0xD800 || !isLeadSurrogate(currentCodePoint)) return;
```
caitp: fast case: U+D7FF is a pretty big range of codes, and is the last code point before a trail of lead surrogates --- slightly slower are at the higher end of the BMP, which are detected with the function call

src/syntax/Scanner.js
```diff
@@ -1172,11 +1172,33 @@ function isAtEnd() {
...
+  let nextCharCode = input.charCodeAt(index + 1) | 0;
+  // ASSERT: isTrailSurrogate(nextCharCode)
+  // Should report an error if it's not?
```
caitp: could get rid of `isTrailSurrogate()` if we never care about this, I guess

arv:
@caitp Same Q here. Do you want to continue with this?

caitp:
did we decide it was worth it? it's a performance hit for probably no practical purpose, but you should make the call

arv:
Closing this one too...

PR #1825, opened 2015-03-16
Create table of Id codepoints which are not spec'd for ES5

caitp:


caitp:
Followup for #1824 which will allow traceur to know if it needs to translate an identifier when generating ES5 code

caitp:
/ping @mathiasbynens I got the idea to do these patches from your blog post, does it look roughly right to you? (minus the redundant `(L&)|(Lu)|(Ll)|(Lt)`)

mathiasbynens:
Just some high-level comments for now:

It would be safer to compare ES6 + Unicode v7.0.0 data with ES5 + Unicode v3.0.0 data, i.e. the minimal required version as per the ES5 spec.

Any reason you’re not using https://www.npmjs.com/package/unicode-7.0.0 (and https://www.npmjs.com/package/unicode-5.1.0 and https://www.npmjs.com/package/unicode-3.0.0 etc.)? It would simplify this script greatly — you wouldn’t have to download/parse the `.txt` file anymore. You could just import the data directly:

```js
const ID_Start = require('unicode-7.0.0/properties/ID_Start/code-points');
const ID_Continue = require('unicode-7.0.0/properties/ID_Continue/code-points');
```

I’ll take a closer look at the code asap.

caitp:
yeah, I thought about that but it might be bad to incorporate more dependencies or something, see what the owners think :>

mathiasbynens:
They could be `devDependencies`.

caitp:
the unicode 3.0 thing is a good point, but it's worth finding out what browsers are actually doing --- eg if safari 7 and ie9 is okay with the current heuristics, that's probably good enough for most applications

mathiasbynens:
> if safari 7 and ie9 is okay with the current rules, that's probably good enough for most applications

I link to some relevant bug tickets here: https://mathiasbynens.be/notes/javascript-identifiers#examples

arv:
Having build/build-unicode-tables.js depend on the npm package seems like a great simplification. I wasn't aware of this when I wrote the script.

I think the bigger question about which identifiers we need to transform is that we should transform identifiers that are not supported in ES5+Unicode3.

Another question is how important this really is. For completeness sake I'm in favor of doing this but I don't think anyone really cares for this. Using non ascii identifiers is bad practice in most projects.

caitp:
> Another question is how important this really is. For completeness sake I'm in favor of doing this but I don't think anyone really cares for this. Using non ascii identifiers is bad practice in most projects.

I agree with this --- but I am a fan of completeness, too

arv:
@caitp Are you planning to get back to this? I'm fine closing if not.

arv:
Closing :crying_cat_face:

PR #1834, opened 2015-03-21
Support parenthesized AssignmentPatterns

### LINKED ISSUE #1833: LHS of assignment expressions are never(?) parsed as AssignmentPatterns
The following is reported as a syntax error:

```js
var a, b;
({a, b}) = {a: 1, b: 2};
```

Based on the draft and other implementations, it seems like this is not right:

From the draft:

>In certain circumstances when processing the production AssignmentExpression : LeftHandSideExpression = AssignmentExpression the following grammar is used to refine the interpretation of LeftHandSideExpression.

(particularly, when the LHS looks like an object literal or array literal)

From SpiderMonkey:

```c++
/*
 * Destructuring patterns can appear in two kinds of contexts:
 *
 * - assignment-like: assignment expressions and |for| loop heads.  In
 *   these cases, the patterns' property value positions can be
 *   arbitrary lvalue expressions; the destructuring is just a fancy
 *   assignment.
 *
 * - binding-like: |var| and |let| declarations, functions' formal
 *   parameter lists, |catch| clauses, and comprehension tails.  In
 *   these cases, the patterns' property value positions must be
 *   simple names; the destructuring defines them as new variables.
 *
 * In both cases, other code parses the pattern as an arbitrary
 * primaryExpr, and then, here in checkDestructuring, verify that the
 * tree is a valid AssignmentPattern or BindingPattern.
 *
 * In assignment-like contexts, we parse the pattern with
 * pc->inDeclDestructuring clear, so the lvalue expressions in the
 * pattern are parsed normally.  primaryExpr links variable references
 * into the appropriate use chains; creates placeholder definitions;
 * and so on.  checkDestructuring is called with |data| nullptr (since
 * we won't be binding any new names), and we specialize lvalues as
 * appropriate.
 *
 * In declaration-like contexts, the normal variable reference
 * processing would just be an obstruction, because we're going to
 * define the names that appear in the property value positions as new
 * variables anyway.  In this case, we parse the pattern with
 * pc->inDeclDestructuring set, which directs primaryExpr to leave
 * whatever name nodes it creates unconnected.  Then, here in
 * checkDestructuring, we require the pattern's property value
 * positions to be simple names, and define them as appropriate to the
 * context.  For these calls, |data| points to the right sort of
 * BindData.
 */
```

(originally reported in https://github.com/jshint/jshint/issues/2269)

caitp:
Closes #1833

### DIFF
--- a/src/syntax/Parser.js
+++ b/src/syntax/Parser.js
@@ -24,6 +24,7 @@ import {
   IDENTIFIER_EXPRESSION,
   LITERAL_PROPERTY_NAME,
   OBJECT_LITERAL_EXPRESSION,
+  PAREN_EXPRESSION,
   REST_PARAMETER,
   SYNTAX_ERROR_TREE
 } from './trees/ParseTreeType.js';
@@ -2788,6 +2789,13 @@ export class Parser {
         // If we fail to parse as an AssignmentPattern then
         // parseAssignmentPattern_ will take care reporting errors.
         return this.parseAssignmentPattern_();
+      case PAREN_EXPRESSION:
+        let location = tree.expression.location;
+        tree = this.transformLeftHandSideExpression_(tree.expression);
+        if (tree.isPattern()) {
+          this.scanner_.index = location.end.offset + 1;
+        }
+        return new ParenExpression(location, tree);
     }
     return tree;
   }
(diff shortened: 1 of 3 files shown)


caitp:
Closing based on https://github.com/jshint/jshint/issues/2269#issuecomment-84515787 for now

PR #1839, opened 2015-03-25
readme: AtScript no longer exists

randombk:
Rename "AtScript" to "TypeScript", as the AtScript name is now retired. All AtScript functionality is being moved to TypeScript.

Signed-off-by: David Li <jiawei.davidli@gmail.com>


arv:
I would call these Angular extensions ;-)

randombk:
That is what the functionality is being used for at the moment, but it looks like most of the extensions are going to be merged into TypeScript, opening them up for use in other places.

arv:
I updated this.

PR #1844, opened 2015-03-28
[strong] Fix issue with placeholder parser

arv:
This allows us to use the template literal supported by V8/iojs


arv:
@domenic PTAL

With the old incorrect caching behavior (#1845) we did not share the tree between

```js
parseStatement `x`
```
and
```js
parseStatements `x`
```

but with the correct template behavior we end up with the same tree.

domenic:
Makes sense in light of #1845.

PR #1849, opened 2015-03-29
Revert "Remove unused module getopt"

arv:
This reverts commit bb7c8e3ee52245d5d401cda017f4d017263dca73.

getopt is being used by build/expand-js-template.js. We need to
replace that before this can be removed.


arv:
@johnjbarton TBR

PR #1858, opened 2015-04-01
fixes windows paths in node tests

pflannery:
fixes source map relative path test for windows

### DIFF
--- a/test/unit/node/generated-code-dependencies.js
+++ b/test/unit/node/generated-code-dependencies.js
@@ -272,13 +272,10 @@ suite('context test', function() {
   });
 
   test('sourceMappingURL from --source-maps=true and deep directories', function(done){
-    // FIXME: exec fails to execute traceur on windows.
-    if (process.platform === 'win32') return done();
-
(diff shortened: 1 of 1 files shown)


domenic:
This seems to depend on Traceur being installed globally?

arv:
What @domenic said. We want the test to test the current version of traceur and not the installed one.

I believe the solution is to use `node ../../traceur` but I ran into some issues with that.

pflannery:
> This seems to depend on Traceur being installed globally?
@domenic your right it was, I realised that just after I went out. I've fixed now.

@arv I've got it to work with the original paths. it seemed like this was the only [failing point](https://github.com/pflannery/traceur-compiler/blob/1a4f752430d2d2282486da7470316ba2a0ffcb31/test/unit/node/generated-code-dependencies.js#L303) but the '\' replace solves that too.

Sorry that I had to bring this back up but it's got in the way of the other PR where I'm trying to solve another route other than use options for the `rootModules` property...

arv:
LGTM

PR #1871, opened 2015-04-07
Remove getForTesting

arv:


arv:
@johnjbarton PTAL

test/unit/browser/System.js
```diff
@@ -83,7 +83,7 @@ suite('System.js', function() {
 
   test('System.set', function() {
     var store = $traceurRuntime.ModuleStore;
-    var polyfills = store.getForTesting('src/runtime/polyfills/polyfills.js');
+    var polyfills = store.get(System.version + '/src/runtime/polyfills/polyfills.js');
```
arv: I could not figure out how to test this one. It gets reference from test/runtime.html but that is completely broken.

johnjbarton: Did you mean "It gets referenced from test/runtime.html, a file that does not work now"?

johnjbarton: We could add 'test/unit/browser/*.js' to the patterns, then remove it in NodeTraceurTestRunner. 

I should have done the `/node/` exclusion in BrowserTraceurTestRunner.

arv: https://github.com/google/traceur-compiler/blob/master/test/runtime.html#L13

It just uses a script[src] to an es6 file. This probably broke a while ago. We should probably remover this html file and only use 1 html test runner.

johnjbarton: The purpose of that test was to verify `../bin/traceur-runtime.js`in the
browser. But since we did not run it, it does nothing.  Removing is fine by
me.

arv:
@johnjbarton Ping

johnjbarton:
LGTM

PR #1877, opened 2015-04-10
[strong] Strongify files in build/ which are passed to node

marjaholtta:
... these ones are compiled w/ node when building Traceur and should be safe to strongify, afaics...


marjaholtta:
Huh, I'm failing with pull requests somehow; this does not look correct. I'd like to merge from my branch into the strong-mode branch but this doesn't look like it requests the right thing.

PR #1878, opened 2015-04-10
[strong] Strongify files in build/ which are passed to node during the b...

marjaholtta:
...uild.


marjaholtta:
These ones are compiled w/ node when building Traceur and should be safe to strongify, afaics.

arv:
LGTM

PR #1896, opened 2015-04-21
[strong] Disallow assigning to undefined and binding

arv:
In strong mode it is not allowed to create a new binding with the
name undefined. It is also not allowed to assign to the existing
undefined.


arv:
@rossberg-chromium PTAL

PR #1900, opened 2015-04-22
Make TempVarTransformer use let where possible

arv:
If the generated code outputs let then make the TempVarTransformer
use let bindings instead of vars.

### DIFF
--- a/src/codegeneration/TempVarTransformer.js
+++ b/src/codegeneration/TempVarTransformer.js
@@ -21,7 +21,7 @@ import {
 } from '../syntax/trees/ParseTrees.js';
 import {ARGUMENTS} from '../syntax/PredefinedName.js';
 import {StringSet} from '../util/StringSet.js';
-import {VAR} from '../syntax/TokenType.js';
+import {LET, VAR} from '../syntax/TokenType.js';
 import {
   createFunctionBody,
   createThisExpression,
@@ -60,10 +60,13 @@ class TempScope {
 }
 
 class VarScope {
-  constructor() {
+  constructor(options) {
     this.thisName = null;
     this.argumentName = null;
     this.tempVarStatements = [];
+    this.declarationType_ =
+        options.blockBinding && !options.transformOptions.blockBinding ?
+            LET : VAR;
   }
 
   push(tempVarStatement) {
@@ -99,7 +102,7 @@ class VarScope {
     }
 
     return createVariableStatement(
-        createVariableDeclarationList(VAR, declarations));
+        createVariableDeclarationList(this.declarationType_, declarations));
   }
 }
 
@@ -110,13 +113,17 @@ class VarScope {
 export class TempVarTransformer extends ParseTreeTransformer {
   /**
    * @param {UniqueIdentifierGenerator} identifierGenerator
+   * @param {ErrorReporter} reporter
+   * @param {Options} options
    */
-  constructor(identifierGenerator) {
+  constructor(identifierGenerator, reporter, options) {
     super();
     this.identifierGenerator = identifierGenerator;
+    this.reporter = reporter;
+    this.options = options;
 
     // Stack used for variable declarations.
-    this.tempVarStack_ = [new VarScope()];
+    this.tempVarStack_ = [new VarScope(this.options)];
 
     // Stack used for the temp scopes. Temp scopes can be used to allow
     // temporary names to used sooner than after leaving the currently function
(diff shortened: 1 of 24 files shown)


arv:
@johnjbarton PTAL

Even though this PR is for the strong mode branch I plan to merge this commi onto master. The main reason for developing this on the strong mode branch is that it TDZ works better in V8 than in Traceur so it would have caught possible issues with using let.

test/unit/codegeneration/TempVarTransformerUseLet.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  testResult(USE_VAR, '({a, b} = {})',
+      'var $__0;\n' +
+      '(($__0 = {}, a = $__0.a, b = $__0.b, $__0));\n' +
+      '//# sourceURL=<compileSource>\n');
+  testResult(USE_LET, '({a, b} = {})',
+      'let $__0;\n' +
+      '(($__0 = {}, a = $__0.a, b = $__0.b, $__0));\n' +
+      '//# sourceURL=<compileSource>\n');
```
arv: @johnjbarton How do I get rid of this? I passed `sourceMaps: false` above already.

johnjbarton: Which "this" do you want to get rid of ? sourceURL?

johnjbarton:
```
let result = compiler.compile(content, 'test.js', 'test.js');
```
Should omit the sourceURL.  (Seems silly I know. I think we should go back to separate the debug info from the compile step.)

arv: Yup. The `//# sourceURL` comment.

johnjbarton:
I'm unsure if you want to try my output name == input name suggestion or shall I just review as is?

arv:
I'm ok with the extra cruft at this time.

PTAL

src/codegeneration/TempVarTransformer.js
```diff
@@ -110,13 +113,15 @@ class VarScope {
 export class TempVarTransformer extends ParseTreeTransformer {
   /**
    * @param {UniqueIdentifierGenerator} identifierGenerator
+   * @param {Options} options
    */
-  constructor(identifierGenerator) {
+  constructor(identifierGenerator, options) {
```
johnjbarton: I guess the other way to go here is to pass `reporter` in to `TempVarTransformer` so all of the transformers use the same constructor parameters. That way we don't need the dummy constructors that only shorten the arg list. The unused reporter argument is common to many transformers, they don't emit errors at this time.

test/unit/codegeneration/TempVarTransformerUseLet.js
```diff
@@ -0,0 +1,60 @@
...
+
+  function testResult(blockBinding, content, expectedResult) {
+    test('', function() {
+      let compiler = new Compiler({
+        blockBinding,
+        script: true,
+        sourceMaps: false
+      });
+      let result = compiler.compile(content);
+      assert.equal(result, expectedResult);
```
johnjbarton: Maybe we should just check the first three characters? The test should verify that we emitted `let`/'var` when we should emit `let`/`var`.  If the test fails because of a mismatch in other parts of the expected output, it won't be because of `TempVarTransformer`.

johnjbarton:
LGTM

src/codegeneration/TempVarTransformer.js
arv: I contemplated doing this. My opinion was that it is cleaner to only pass what is needed but after doing all this work I think I've changed my mind. Let me make ParseTreeTransformer take three params and expose them as properties.

test/unit/codegeneration/TempVarTransformerUseLet.js
arv: Good idea. I can also do a visitor that would throw if I find a the wrong kind of declaration.

src/codegeneration/TempVarTransformer.js
arv: Making ParseTreeTransformer take `identifierGenerator, reporter, options` caused a lot of churn. Gave up on that. I'll try to TempVarTransformer now.

arv:
Updated the test to use a visitor and made TempVarTransformer take reporter too.

PTAL

PR #1903, opened 2015-04-25
Fix issue with prepend statement

### LINKED ISSUE #1199: Strange behavior for parameter transformations

arv:
In case the function only had a directive then the statements did not
get added.

Fixes #1199


arv:
@johnjbarton TBR

I extracted this in advance of #1902

PR #1904, opened 2015-04-25
Move parameter override to ParameterTransformer

arv:
At the moment all subclasses override it to not transform type
annotations but the ParameterTransformer should never transform
parameters in type annotations.


arv:
@johnjbarton TBR

I extracted this too from #1902

PR #1905, opened 2015-04-25
[strong] Fix issue with making strong functions strict

arv:
The transformer was not calling the super class so we were not
correctly tranforming the sub trees.


arv:
@rossberg-chromium TBR

PR #1914, opened 2015-05-03
Ensure export star does not export default

guybedford:
Export * is not supposed to export the default export.

> 15.2.1.16.2 GetExportedNames( exportStarSet ) Concrete Method

>    For each ExportEntry Record e in module.[[StarExportEntries]], do
    Let requestedModule be HostResolveImportedModule(module, e.[[ModuleRequest]]).
    ReturnIfAbrupt(requestedModule).
    Let starNames be requestedModule.GetExportedNames(exportStarSet).
    For each element n of starNames, do
    If SameValue(n, "default") is false, then
    If n is not an element of exportedNames, then
    Append n to exportedNames.

> 15.2.1.16.3 ResolveExport( exportName, resolveSet, exportStarSet ) Concrete Method

>    If SameValue(exportName, "default") is true, then
    Assert: A default export was not explicitly defined by this module.
    Throw a SyntaxError exception.
    NOTE A default export cannot be provided by an export *.

### DIFF
--- a/src/codegeneration/InstantiateModuleTransformer.js
+++ b/src/codegeneration/InstantiateModuleTransformer.js
@@ -369,7 +369,7 @@ export class InstantiateModuleTransformer extends ModuleTransformer {
       if (exportStarBinding) {
         setterStatements = setterStatements.concat(parseStatements `
           Object.keys($__m).forEach(function(p) {
-            if (!$__exportNames[p])
+            if (p !== 'default' && !$__exportNames[p])
               $__export(p, $__m[p]);
           });
         `);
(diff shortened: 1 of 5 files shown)


arv:
Can you add a feature test too?

guybedford:
Can you clarify what you'd need? That test case does provide full coverage.

arv:
I don't see how it fixes the same issue for the other module formats?

We prefer writing tests in `test/feature/` if possible since they are easier to maintain.

guybedford:
Sure, I've updated it to cover the other module formats as well.

arv:
Thanks you. LGTM

PR #1925, opened 2015-05-06
Remove all implicit conversions from traceur

arv:


arv:
@johnjbarton TBR

This is a cherry-pick from the strong-mode branch.

PR #1931, opened 2015-05-11
[strong] Assert no type conversion for +=

arv:
Both sides must be string or number.


arv:
@rossberg-chromium PTAL

I forgot this one before.

rossberg:
LGTM

PR #1957, opened 2015-06-10
Implement bulk re-exports for instantiate setter perf

guybedford:
Aurelia had some issues with bulk exports causing performance bottlenecks due to the repeated bindings being called in quadratic multiples of the depth of `export *` usage. This function flattens the binding call in one go. It may be worth always having `$__export({ name: 'value' })` as an object so we don't need to overload the function, so could look at a PR to do this if you think it is worthwhile, but the named form of `$__export(name, 'value')` will still be necessary for backwards compatibility for a while regardless.

Associated ModuleLoader PR is at https://github.com/ModuleLoader/es6-module-loader/pull/387.

### DIFF
--- a/src/codegeneration/InstantiateModuleTransformer.js
+++ b/src/codegeneration/InstantiateModuleTransformer.js
@@ -349,28 +349,26 @@ export class InstantiateModuleTransformer extends ModuleTransformer {
 
       // then do export bindings of re-exported dependencies
       if (externalExportBindings) {
-        let reexports = {};
+        let reexports = Object.create(null);
         externalExportBindings.forEach(({exportName, importName}) => {
           reexports[exportName] = importName === null ?
               parseExpression `$__m` : parseExpression `$__m.${importName}`;
         });
         setterStatements.push(
-          parseStatement `$__export(${createObjectLiteral(reexports)})`
-        );
+            parseStatement `$__export(${createObjectLiteral(reexports)})`);
       }
 
       // create local module bindings
       if (moduleBinding) {
         setterStatements.push(
-          parseStatement `${id(moduleBinding)} = $__m;`
-        );
+            parseStatement `${id(moduleBinding)} = $__m;`);
       }
 
       // finally run export * if applying to this dependency, for not-already
       // exported dependencies
       if (exportStarBinding) {
         setterStatements = setterStatements.concat(parseStatements `
-          var exportObj = {};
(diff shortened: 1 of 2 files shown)


src/codegeneration/InstantiateModuleTransformer.js
```diff
@@ -349,12 +349,14 @@ export class InstantiateModuleTransformer extends ModuleTransformer {
 
       // then do export bindings of re-exported dependencies
       if (externalExportBindings) {
+        let reexports = {};
```
arv: What about `__proto__` etc? You probably need to use `Object.create(null)` or `new StringMap()`.

src/codegeneration/InstantiateModuleTransformer.js
```diff
@@ -349,12 +349,14 @@ export class InstantiateModuleTransformer extends ModuleTransformer {
...
         externalExportBindings.forEach(({exportName, importName}) => {
-          let statement = importName === null ?
-              parseStatement `$__export(${exportName}, $__m);` :
-              parseStatement `$__export(${exportName}, $__m.${importName});`;
-          setterStatements.push(statement);
+          reexports[exportName] = importName === null ?
+              parseExpression `$__m` : parseExpression `$__m.${importName}`;
         });
+        setterStatements.push(
+          parseStatement `$__export(${createObjectLiteral(reexports)})`
```
arv: +2 spaces indent

src/codegeneration/InstantiateModuleTransformer.js
```diff
@@ -368,10 +370,12 @@ export class InstantiateModuleTransformer extends ModuleTransformer {
       // exported dependencies
       if (exportStarBinding) {
         setterStatements = setterStatements.concat(parseStatements `
+          var exportObj = {};
```
arv: same issue here with obejcts as maps

arv: Yeah, this one needs to be `Object.create(null)`. If not a, an export named `__proto__` will screw everything up.

arv:
Conceptually this looks good.

guybedford:
Thanks, I've updated the PR.

arv:
LGTM, I'll squash and merge

arv:
Committed as 84956a25a6b05b0659b961e27d8a0313fd3372cf

Issue #1997, opened 2015-10-29
import failures without the .js postfix


iislucas:
Given: 

test.js: 
```
import * as toylib from "./toylib";
toylib.toy();
```

toylib.js:
```
export function toy() {
    console.log("toy library called.")
    return "toy!";
}
```

```
$ traceur test.js --out test_es5.js
Error: Error: ENOENT: no such file or directory, open '/Users/ldixon/code/respect-engine/typescript-polymer/build/home_page/toylib'
Specified as ./toylib.
Imported by test.js.
Normalizes to toylib
locate resolved against base '/Users/ldixon/code/respect-engine/typescript-polymer/build/home_page/'

    at InternalLoader.loaderCompiler.handleCodeUnitLoadError (/Users/ldixon/code/respect-engine/typescript-polymer/node_modules/traceur/bin/traceur.js:31014:40)
    at /Users/ldixon/code/respect-engine/typescript-polymer/node_modules/traceur/bin/traceur.js:30872:21
```

However, if I change the import statement to: `import * as toylib from "./toylib.js"`, then it compiles as expected. 

I think ES6 is supposed to correctly guess `.js` extensions (from looking at places like: http://www.2ality.com/2014/09/es6-modules-final.html and https://github.com/ModuleLoader/es6-module-loader/wiki/Brief-Overview-of-ES6-Module-syntax and http://www.ecma-international.org/ecma-262/6.0/index.html#table-40 )

iislucas:
OK, Having looked around the Traceur code, I think I see what's happening... 
https://github.com/google/traceur-compiler/blob/master/src/runtime/TraceurLoader.js#L143

The fetch call assumes the address given is a fully resolved address, but when you miss the .js postfix, this is false. There are several levels this can be fixed at: 

1. the loader implementation, e.g. https://github.com/google/traceur-compiler/blob/master/src/node/nodeLoader.js  

2. The Traveur loader: https://github.com/google/traceur-compiler/blob/master/src/runtime/TraceurLoader.js#L143 If the original fetch fails, we can try a new address with postfixed ".js" in the fetch. 

(2) looks like it's probably the best way to go. I'll put together a pull request to experiment with this fix.

arv:
This is intentional.

We made this change to better match the upcoming spec.

iislucas:
I see. In that case I need to file a bug in the compiler that is producing the import statements that miss the `.js` postfix. Can you point me to where the es6 spec says that the import names should include the postfix? 
[...]

PR #1999, opened 2015-10-30
Make compatible with TypeScript compiler (by adding ".js" postfix if it was missing)

iislucas:
If a module fails to load, and the module name does not end in ".js", and adding ".js" postfix will make the file load, then do that. This makes traceur compatible with TypeScript, which current emits import statements that don't include the ".js" postfix. 

See: 
https://github.com/google/traceur-compiler/issues/1997
and 
https://github.com/Microsoft/TypeScript/issues/5460

Added test, and testes with `make test`

[Review on Reviewable](https://reviewable.io/reviews/google/traceur-compiler/1999)

### DIFF
--- a/src/runtime/TraceurLoader.js
+++ b/src/runtime/TraceurLoader.js
@@ -141,14 +141,24 @@ export class TraceurLoader extends Loader {
   }
 
   fetch(load) {
-    return new Promise((resolve, reject) => {
-      if (!load)
-        reject(new TypeError('fetch requires argument object'));
-      else if (!load.address || typeof load.address !== 'string')
-        reject(new TypeError('fetch({address}) missing required string.'));
+    if (!load)
+      return Promise.reject(new TypeError('fetch requires argument object'));
+
+    if (!load.address || typeof load.address !== 'string')
+      return Promise.reject(new TypeError(
+          'fetch({address}) missing required string.'));
+
+    return (new Promise((resolve, reject) => {
+      this.fileLoader_.load(load.address, resolve, reject);
+    }))
+    .catch((e) => {
+      if (load.address.slice(-3) !== '.js')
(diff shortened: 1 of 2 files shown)


arv:
I think this needs to be guarded by an option. We don't want to go back to allowing this by default.


---

Reviewed 2 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

<sup>**[src/runtime/TraceurLoader.js, line 155 \[r1\]](https://reviewable.io:443/reviews/google/traceur-compiler/1999#-K1tGKhS9IneJtgzC7QM)** ([raw file](https://github.com/google/traceur-compiler/blob/791cbf1f448daecea8438cd6c1275e7020245f24/src/runtime/TraceurLoader.js#L155)):</sup>
Missing `{}`


---

<sup>**[test/unit/runtime/Loader.js, line 150 \[r1\]](https://reviewable.io:443/reviews/google/traceur-compiler/1999#-K1tGaZUPi1b5HTr89et)** ([raw file](https://github.com/google/traceur-compiler/blob/791cbf1f448daecea8438cd6c1275e7020245f24/test/unit/runtime/Loader.js#L150)):</sup>
You can use ES6 here so a template literal would have been cleaner.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/traceur-compiler/1999)

iislucas:
Just closing the loop on this. I ended up finding another way to do this, so I don't mind leaving this as is. If someone else needs traceur, then can easily finish this up. :)

arv:
OOC did TypeScript finally change this?

iislucas:
I don't think so; I just found an easier way using the closure compiler, which solves another problem I was having also (actually doing the module loading).

PR #2010, opened 2015-11-17
Add modules ignore option

guybedford:
I'm in the process of splitting SystemJS out so that transpilation always happens through plugins and not as part of SystemJS core itself.

In order to construct a Traceur plugin that can optimize ES modules nicely for builds, the ability to ignore module syntax entirely would be incredibly useful.

### DIFF
--- a/src/Options.js
+++ b/src/Options.js
@@ -91,7 +91,7 @@ let defaultValues = Object.create(null);
 let featureOptions = Object.create(null);
 let experimentalOptions = Object.create(null);
 let moduleOptions =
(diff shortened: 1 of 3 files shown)


arv:
`--modules-parse` used to work to do this but it broke over the years.

Can you rename `ignore` to `parse`? Our options are covering both parsing and transforming and `ignore` does not make it clear that parsing is supported.

Other than that, this is a very useful feature.

guybedford:
Thanks for the quick response, sure I've renamed this to `parse`.

arv:
LGTM

guybedford:
Thanks for the quick review!

PR #2020, opened 2015-11-21
Add abstraction for private symbols

arv:
Add the following runtime functions:

- getPrivate
- setPrivate
- hasPrivate
- deletePrivate

(These are all own property)

The motivation is that we can then use a WeakMap if it is natively
supported.


arv:
@johnjbarton Just the last commit... There are already PRs for the previous commits.

PR #2021, opened 2015-11-23
Use WeakMap for private state

arv:


arv:
@johnjbarton One more... only the last commit as usual.

arv:
I was planning to pull this out into a module. I can pull it out into two different modules and test them individually.

PR #2025, opened 2015-11-25
Add node 0.10 to TravisCI and fix test

arv:


arv:
Something change in how sort compares objects (?).

Removed the sort which is also better since then the correct order is also tested,

arv:
@johnjbarton TBR

PR #2043, opened 2015-12-14
Implement __moduleName second argument to System.register

guybedford:
See https://github.com/systemjs/builder/issues/416 for the discussion here.

### DIFF
--- a/src/codegeneration/InstantiateModuleTransformer.js
+++ b/src/codegeneration/InstantiateModuleTransformer.js
@@ -268,12 +268,12 @@ export class InstantiateModuleTransformer extends ModuleTransformer {
 
     if (this.moduleName) {
       return parseStatements `System.register(${this.moduleName},
-          ${this.dependencies}, function($__export) {
+          ${this.dependencies}, function($__export, __moduleName) {
             ${statements}
           });`;
     } else {
       return parseStatements
-        `System.register(${this.dependencies}, function($__export) {
+        `System.register(${this.dependencies}, function($__export, __moduleName) {
           ${statements}
         });`;
     }
--- a/src/codegeneration/InstantiateModuleTransformer.js
+++ b/src/codegeneration/InstantiateModuleTransformer.js
@@ -214,6 +214,9 @@ export class InstantiateModuleTransformer extends ModuleTransformer {
     this.anonymousModule =
         options && !options.bundle && options.moduleName !== true;
 
+    // does the module reference __moduleName
+    this.usesModuleName = false;
+
     this.inExport_ = false;
     this.curDepIndex_ = null;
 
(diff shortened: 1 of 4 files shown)


arv:
Can this be tested?

becky7:
HI IS BECKY

guybedford:
Sure, I've included a test.

guybedford:
I've also added an adjustment to only output the `__moduleName` argument when it is needed.

src/codegeneration/InstantiateModuleTransformer.js
```diff
@@ -250,6 +253,18 @@ export class InstantiateModuleTransformer extends ModuleTransformer {
     this.exportStarBindings = [];
   }
 
+  transformIdentifierExpression(tree) {
+    if (tree.identifierToken.value == '__moduleName')
+      this.usesModuleName = true;
+    return super.transformIdentifierExpression(tree);
+  }
+
+  transformBindingIdentifier(tree) {
```
arv: This should not be overridden. BindingIdentifiers introduce new bindings... Like so:

```js
function x() {  // x: BindingIdentifier
  var y = 42;  // y: BindingIdentifier
  return z;  // z: IdentifierExpression
}
```

It is enough to override `transformIdentifierExpression` even though it will give a few false positives (which are harmless in this case).

src/codegeneration/InstantiateModuleTransformer.js
```diff
@@ -250,6 +253,18 @@ export class InstantiateModuleTransformer extends ModuleTransformer {
...
+    if (tree.identifierToken.value == '__moduleName')
```
arv: `===`

src/codegeneration/InstantiateModuleTransformer.js
```diff
@@ -250,6 +253,18 @@ export class InstantiateModuleTransformer extends ModuleTransformer {
     this.exportStarBindings = [];
   }
 
+  transformIdentifierExpression(tree) {
+    if (tree.identifierToken.value == '__moduleName')
+      this.usesModuleName = true;
```
arv: `===`

src/codegeneration/InstantiateModuleTransformer.js
```diff
@@ -266,16 +281,30 @@ export class InstantiateModuleTransformer extends ModuleTransformer {
...
-    if (this.moduleName) {
-      return parseStatements `System.register(${this.moduleName},
-          ${this.dependencies}, function($__export) {
+    if (this.usesModuleName) {
+      if (this.moduleName) {
+        return parseStatements `System.register(${this.moduleName},
+            ${this.dependencies}, function($__export, __moduleName) {
+              ${statements}
+            });`;
+      } else {
```
arv: skip else after return

test/instantiate/module-name.js
```diff
@@ -0,0 +1 @@
+export var name = __moduleName;
```
arv: Can this file also include:

```js
export function f(__moduleName) {
  return __moduleName;
}
```

and then test that `f(42) === 42`

arv:
Overall looking good. Just a few nits and it is good to go.

guybedford:
Sure thanks, I've updated the PR.

arv:
Committed as 03605414e9c97d6b69e01365c2b8ea4fae48744c

PR #2065, opened 2016-01-18
Compile src to dist as commonjs modules.

arv:
Should we use `dist/commonjs` or maybe `src/commonjs`? Right now I'm using `dist` which is not that nice since we are all moving to an ES module world and commonjs modules will hopefully go away.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -299,6 +299,9 @@ updateSemver: # unless the package.json has been manually edited.
 	node build/versionInfo.js -v > build/npm-version-number
 	git diff --quiet -- package.json && node build/incrementSemver.js
 
+dist/commonjs:
+	./traceur --dir src/ dist/commonjs/ --modules=commonjs
+
 # --- Targets that push upstream.
 
 # We start with a clean repo and an 'upstream' remote like github sets up.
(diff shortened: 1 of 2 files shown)


arv:
@johnjbarton I can squash all these PRs into one if you prefer?

@guybedford Does this solve your issue?

arv:
Another thing to consider is to require Node 5 and disable transformations that are natively supported there.

arv:
One more option would be to publish these as a different npm package... but I don't like that... Do you?

arv:
@johnjbarton PTAL

I changed it to compile to `dist/commonjs` since that leaves us more room for other output formats in the future.

johnjbarton:
yes I think the dist version is better, LGTM

arv:
@johnjbarton Your content is a bit unclear to me; "dist version" as in dist/ or "dist version" as in dist/commonjs/?

johnjbarton:
sorry, I meant your last change, with the sub-dir.

arv:
@johnjbarton I think I need to update the npm script version too... We should remove the make version once we resolved the current snafus.

PR #2066, opened 2016-01-19
Fix spacing after period

jshen212:
Fixed spacing after period on line 100


arv:
LGTM

But we you need to sign the CLA and add yourself to AUTHORS which I understand it might not be worth doing. See https://github.com/google/traceur-compiler/blob/master/CONTRIBUTING.md for more details.

PR #2069, opened 2016-01-25
Update Options.js

danielrussellLA:
cleaned up various comments.

### DIFF
--- a/src/Options.js
+++ b/src/Options.js
@@ -71,7 +71,7 @@ export const optionsV01 = enumerableOnlyObject({
 
 export const versionLockedOptions = optionsV01;
 
-// Options are just a plain old object. There are two read only views on this
+// Options are just a plain old object. There are two read-only views on this
 // object, parseOptions and transformOptions.
 //
 // To set an option you do `options.classes = true`.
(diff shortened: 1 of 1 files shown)


arv:
Thanks for fixing all the comments.

To accept this you also have to sign the Google CLA and add yourself to AUTHORS. See https://github.com/google/traceur-compiler/blob/master/CONTRIBUTING.md for more details.

Assigning to @johnjbarton since only Googlers can verify that you have signed the CLA.

danielrussellLA:
Done. Thank you Erik!

src/Options.js
```diff
@@ -487,7 +487,7 @@ export function toDashCase(s) {
  * TODO(jjb): move to src/node
  * This is called by build.js to add options to the commander command line
  * library.
- * @param {Commander} flags The commander object.
+ * @param {Commander} flags the commander object.
```
johnjbarton: The word after the type is the argument name. This change makes it read as if the word `flags` is a verb operating on `the commander object`.  Perhaps `flags: the commander object` or `flags, the commander object`.

src/Options.js
```diff
@@ -554,7 +554,7 @@ export function addOptions(flags, commandOptions) {
   Object.keys(commandOptions).forEach(function(name) {
     let dashedName = toDashCase(name);
     if (flags.optionFor('--' + name) || flags.optionFor('--' + dashedName)) {
-      return;   // non-boolean already in flags.
+      return;   // Non-boolean already in flags.
```
johnjbarton: maybe
`These non-boolean options were previously added into flags.`

PR #2132, opened 2016-08-25
Use universal method of finding the global object

cyco:
The module store, runtime and symbol polyfills need access to the global object. This was done by checking the existance of `window`, `global`, `self` and falling back to `this` if none were defined. This can fail for two reasons. First, `this` should be expected to be set to `undefined` during module definitions (as is the case for symbols.js), thus rendering the fallback useless. Secondly some JavaScript implementations (e.g. JavaScript-Core Framework) might not even define a name for the global object that could be accessed within modules.
By using the Function-Constructor we create a new function in non-strict mode which has its _thisArg_ set to the gobal object if no other _thisArg_ is provided.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -56,3 +56,4 @@ Srinivasan Sekar <srinivasan.sekar1990@gmail.com>
 Jeff Shen<jeffshen86@gmail.com>
 Erik Arvidsson <erik.arvidsson@gmail.com>
 Mathias Rangel Wulff <m@rawu.dk>
+Christoph Leimbrock <christoph.leimbrock@gmx.de>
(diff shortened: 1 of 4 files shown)


cyco:
I signed it!

arv:
Thank you for the patch.

One issue with this approach is that it does not work when eval is disabled (which is the case of CSP). So it needs to be wrapped in a try/catch with a fallback to the current ugly behavior.

cyco:
Thanks for your feedback. I'll add the change you suggested and update the PR.

arv:
I like it 👍

Can you also add an entry to AUTHORS?

cyco:
Done :)

arv:
Can you fix the merge conflict? Thanks.

cyco:
The conflict should be resolved now.

PR #2133, opened 2016-09-10
Correct check for undefined variable

mathiasrw:
The code was comparing the result of `typeof` with a variable named `undefined`.

As typeof returns a string it should compare to `'undefined'`

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -55,3 +55,4 @@ Caitlin Potter <caitpotter88@gmail.com>
 Srinivasan Sekar <srinivasan.sekar1990@gmail.com>
 Jeff Shen<jeffshen86@gmail.com>
 Erik Arvidsson <erik.arvidsson@gmail.com>
+Mathias Rangel Wulff <m@rawu.dk>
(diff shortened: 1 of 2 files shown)


arv:
Thank you.

I'll merge once the tests pass.

arv:
@mathiasrw Can you also add an entry to AUTHORS?

mathiasrw:
Sure :)

arv:
Thanks.

This is consistently failing on `node v6.5.0 (npm v3.10.3)`. Not sure why that is?

mathiasrw:
Interesting. 

I will look into if node 6.5 have any special things regarding `typeof`

mathiasrw:
@arv 

When testing the master branch - without my changed code - it works on node 6.4 but not on 6.5: https://travis-ci.org/mathiasrw/traceur-compiler/builds/159081592

Node 6.5 was released the 8th of September: https://nodejs.org/en/download/releases/

I see that last test run was 9th of June: https://travis-ci.org/google/traceur-compiler/builds

So, the test fails because of a new Node release...

I will try to dig a bit down into why and come back with inputs if I find anything...

arv:
That's unlucky. I'll look too.

arv:
I fixed the test. Merging.

Thanks again.

mathiasrw:
Perfect - thanks for quick action :)

PR #2134, opened 2016-09-11
Fix PropertyMethodAssignment test

arv:
The test was testing the name of a method. With newer versions of
Node, Node uses the ES6 semantics for inferring the function name
which leads to test failures.


arv:
@johnjbarton TBR

@mathiasrw FYI

